PR #98, opened 2014-12-03
Fill options value if /proc/cmdline is empty

k0da:
In case installer have empty /proc/cmdline set options value to ""
Otherwise it would fail. bsc#908185

Signed-off-by: Dinar Valeev <dvaleev@suse.com>


lslezak:
Closing without merging, according to IRC discussion this change is actually not needed.

PR #99, opened 2015-01-22
resolve shortcut conflict in repository type dialog (bnc#913099)

lslezak:
to avoid different auto correction in Qt and ncurses UI

- 3.1.56

### DIFF
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -42,7 +42,7 @@ module Yast
       # radio button
       :https             => N_("HTT&PS..."),
       # radio button
-      :samba             => N_("&SMB/CIFS"),
+      :samba             => N_("S&MB/CIFS"),
       # radio button
       :nfs               => N_("NF&S..."),
       # radio button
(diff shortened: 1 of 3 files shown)


src/modules/SourceDialogs.rb
```diff
@@ -44,7 +44,7 @@ class SourceDialogsClass < Module
       # radio button
       :samba             => N_("&SMB/CIFS"),
       # radio button
-      :nfs               => N_("&NFS..."),
+      :nfs               => N_("NF&S..."),
```
mvidner: Nice try, but it doesn't work, does it?

```ruby
:ftp => N_("&FTP..."),
:samba => N_("&SMB/CIFS"),
:next => N_("&Next"), # implied
:nfs => N_("NF&S..."),
```

lslezak: Ha, I overlooked this one... But strangely enough it works when I tested it...

mvidner: ncurses with Y2DEBUG revealed me its logic: it first resolved the conflicts among the radio buttons. Then it assigned the Next button label. Previously, NFS kept its *preferred* key N, so Next lost against it. Now, even though the radio buttons have a conflict, nobody gets N so Next can keep it.

The shortcuts are still different in Qt and ncurses, only Next now gets N.

I think it is acceptable. LGTM.

mvidner:
LGTM

PR #100, opened 2015-01-26
Killing bug 12

jreidinger:


package/yast2-packager.changes
```diff
@@ -1,4 +1,9 @@
 -------------------------------------------------------------------
+Mon Jan 26 13:06:17 UTC 2015 - jreidinger@suse.com
+
+- fixed \r character lost during conversion to Ruby
```
lslezak: NP: in this case there are also `\b` characters...

lslezak:
LGTM

package/yast2-packager.changes
jreidinger: you are right

PR #105, opened 2015-02-03
Handle file:// URL scheme (boo#869399)

lslezak:
- Treat `file://` scheme as `dir://`, it's the same for libzypp (just keep the scheme when editing)
- Validate scheme for a manually entered URL
- Display `/` path if path is empty (for `file` and `dir` scheme)

See https://bugzilla.suse.com/show_bug.cgi?id=869399


cwh42:
LGTM

PR #106, opened 2015-02-03
Use options from partitioner during test mount (bcn#915758)

ancorgs:
This is a backport of the fix at version 3.1.43


lslezak:
Just a note: For backports it's a good idea to mention the original commit or PR so we can easily compare the Ruby and the YCP version.

In this case it's PR #80.

lslezak:
LGTM

lslezak:
BTW, what about PR #81? It is also related to device mounting, maybe it should be backported too?

ancorgs:
Ok, I will include more information about the backport and reuse this PR to also backport PR #81

PR #108, opened 2015-02-03
Use DM device for encrypted partitions for test mount (bnc#889334)

ancorgs:
This is a backport of the fix in version 3.1.44 (pull request #81)


src/modules/SpaceCalculation.ycp
```diff
@@ -642,8 +642,18 @@
 				// join the options
 				string mount_options_str = mergestring(mount_options, ",");
 
+				string device = "";
+				if (part["crypt_device"]:false)
```
lslezak: Use `haskey()` here:  `if (haskey(part, "crypt_device"))`, the default value is used also for type information, it this case the value is not boolean.

See http://doc.opensuse.org/projects/YaST/SLES11/tdg/haskey.html

lslezak:
LGTM

PR #111, opened 2015-02-06
Make the Packages::sourceAccessPackages() function public (bnc#916200)

lslezak:
...to allow using it from the update module

- 2.17.112

Note: This is just a preparation for the fix in the update module...


jreidinger:
LGTM

PR #112, opened 2015-02-09
Use InstLan instead of InstNetworkSetup (bnc#911132)

ancorgs:
Tested on real hardware.

I have a question. This affects the installation process, thus a maintenance update would not be enough. Do we release DUDs, updated ISOs or something similar for openSUSE? I'll create a maintenance request anyways but...


jreidinger:
LGTM

jreidinger:
no we do not release

PR #115, opened 2015-03-03
Fixed addon-license dialog (bnc#899892 & bnc#899891)

cwh42:


src/modules/ProductLicense.rb
```diff
@@ -1144,12 +1144,24 @@ def AskLicenseAgreement(src_id, dir, patterns, action, enable_back, base_product
       end
 
       licenses_ref = arg_ref(licenses)
-      DisplayLicenseDialog(
+
+      label = Pkg::SourceGeneralData(src_id)["name"]
+      title = _("License Agreement")
+
+      if !label.empty?
```
lslezak: Add a comment for translators, describe what the `%s` mean so they can translate it properly.

    # TRANSLATORS: %s is a repository name

src/modules/ProductLicense.rb
```diff
@@ -1144,12 +1144,24 @@ def AskLicenseAgreement(src_id, dir, patterns, action, enable_back, base_product
...
+        title = _("%s License Agreement") % label
+      end
+
+      eula_url = Pkg::SourceGeneralData(src_id)["url"]
+      Yast::ProductLicense.license_file_print = eula_url
```
lslezak: The URL should point to the license file which can be downloaded, actually it is `repository_url` + (`license.zip` or `license.tar.gz`). The used file should be stored in `@license_file_print`, see https://github.com/yast/yast-packager/blob/master/src/modules/ProductLicense.rb#L651.

Moreover you should leave the default location for `cd://` or `dvd://` URLs as there is nothing to "download" in that case.

src/modules/ProductLicense.rb
lslezak: Alternatively you can use `%{}` expansion:

```ruby
repository_name = Pkg::SourceGeneralData(src_id)["name"]
...
_("%{repository_name} License Agreement") % {repository_name: repository_name}
```

In that case the `TRANSLATORS` comment is not needed as the expansion is self descriptive.

src/modules/ProductLicense.rb
cwh42: Since I actually tested my code I found out that the license filename gets appended elsewhere*. So the repo URL is enough here. See screenshot.

* actually in update_license_location() line 1602
addonlicense_fixed

cwh42: And it even checks for location_is_url?

lslezak: The problem is that there is no http://download.opensuse.org/distribution/13.2/repo/oss/license.txt file.

There is http://download.opensuse.org/distribution/13.2/repo/oss/license.tar.gz file which contains license + all translations in one tarball.

cwh42:
I think I found a solution.

src/modules/ProductLicense.rb
```diff
@@ -1601,6 +1613,15 @@ def update_license_location(lang, licenses)
       end
     end
 
+    # update license location displayed in the dialog
+    # @param [Fixnum] src_id integer repository to get the license from.
+    def update_license_archive_location(src_id)
+      src_url = Pkg::SourceGeneralData(src_id)["url"]
+      if location_is_url?(src_url) && UI.WidgetExists(:printing_hint)
+        lic_url = src_url + @license_file_print
```
lslezak: I'd use `File.join` here to properly handle '/' separator when merging the parts.

lslezak:
This cannot be merged, please merge the upstream changes and resolve the conflicts.

cwh42:
Conficts merged & using File.join now.

mvidner:
Please mark bugs as Resolved only once the PR is fixed. I just add a comment in Bugzilla pointing to the PR while it is under review.

mvidner:
LGTM.

PR #121, opened 2015-04-10
Register addons from media (FATE#318505)

lslezak:


package/yast2-packager.spec
```diff
@@ -74,12 +74,10 @@ Requires:	libyui_pkg
...
 
-Provides:	yast2-config-package-manager
-Obsoletes:	yast2-config-package-manager
-Provides:	y2t_spkg y2t_inst-packages y2pkginf y2c_spkg
-Obsoletes:	y2t_spkg y2t_inst-packages y2pkginf y2c_spkg
-Provides:	yast2-trans-package-manager yast2-trans-inst-packages
-Obsoletes:	yast2-trans-package-manager yast2-trans-inst-packages
+# for registering media add-ons
+# (soft dependency because openSUSE does not contain the registration module)
+Recommends:     yast2-registration
```
ancorgs: Couldn't this cause any unwanted warning in openSUSE in some situations? What about using conditionals in the spec file? Just asking, I actually don't know if it's a better solution, I have no clue about packaging.

src/modules/AddOnProduct.rb
```diff
@@ -885,9 +886,15 @@ def RegisterAddOnProduct(src_id)
       # or check the content file
       if WorkflowManager.WorkflowRequiresRegistration(src_id) || Builtins.contains(@addons_requesting_registration, src_id)
         Builtins.y2milestone("Repository ID %1 requests registration", src_id)
-        # TODO FIXME: user needs to manually select the addon to register,
-        # pass the addon so it could be pre-selected
-        WFM.CallFunction("inst_scc", [])
+
+        if !WFM.ClientExists("inst_scc") && !Package.Install("yast2-registration")
```
ancorgs: Having something like the installation of a package kinda "hidden" as the second part of a condition looks strange to me.

test/addon_product_test.rb
```diff
@@ -32,4 +32,59 @@
     end
   end
 
+  describe "#RegisterAddOnProduct" do
+    let(:repo_id) { 42 }
+
+    context "the add-on requires registration" do
+      before do
+        expect(Yast::WorkflowManager).to receive(:WorkflowRequiresRegistration)
```
ancorgs: NP: Semantically '''allow''' in usually more correct in the ```before``` block.

package/yast2-packager.spec
lslezak: Good idea, I'll rather use a macro to not confuse openSUSE...

ancorgs:
lgtm

PR #122, opened 2015-04-23
Force installation of libyui-qt for graphical autoinstallation

ancorgs:
The bug should affect VNC and remote X11 autoyast installations with the minimal pattern (that is, graphical autoinstallations but with X11 not selected for installation).

This was, for example, causing yast to start in ncurses during second stage at https://bugzilla.suse.com/show_bug.cgi?id=923901

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -17,15 +17,15 @@ module Yast
     # All known types of resolvables
     RESOLVABLE_TYPES = [:product, :patch, :package, :pattern, :language]
     # Minimum set of packages required to enable VNC server
-    VNC_BASE_PACKAGES = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts", "xinetd"]
+    VNC_BASE_TAGS = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts", "xinetd"]
     # Additional packages needed to run second stage in graphical mode
     # Use exact package name (i.e. libyui-qt6 instead of libyui-qt) to prevent
     # the UI from asking for confirmation
-    AUTOYAST_X11_PACKAGES = ["libyui-qt6", "yast2-x11"]
+    AUTOYAST_X11_TAGS = ["libyui-qt", "yast2-x11"]
     # Default window manager for VNC if none is installed
     DEFAULT_WM = "icewm"
     # Minimum set of packages required for installation with remote X11 server
-    REMOTE_X11_BASE_PACKAGES = [ "xorg-x11-server", "xorg-x11-fonts", "icewm" ]
+    REMOTE_X11_BASE_TAGS = [ "xorg-x11-server", "xorg-x11-fonts", "icewm" ]
 
     def main
       Yast.import "UI"
@@ -61,6 +61,7 @@ module Yast
       Yast.import "Installation"
       Yast.import "URL"
       Yast.import "PackagesProposal"
+      Yast.import "PackageSystem"
       Yast.import "HTML"
 
       Yast.include self, "packager/load_release_notes.rb"
@@ -138,6 +139,8 @@ module Yast
       @base_source_id = nil
 
       @old_packages_proposal = nil
+
+      PackageSystem.EnsureSourceInit
     end
 
     # summary functions
@@ -2559,12 +2562,12 @@ module Yast
     #
     # @return [Array<String>] package list
     def vnc_packages
-      packages = VNC_BASE_PACKAGES.dup
+      tags = VNC_BASE_TAGS.dup
       # At least one windowmanager must be installed (#427044)
       # If none is there, use a fallback
-      packages << DEFAULT_WM unless has_window_manager?
-      packages.concat(AUTOYAST_X11_PACKAGES) if Mode.autoinst
-      packages
+      tags << DEFAULT_WM unless has_window_manager?
+      tags.concat(AUTOYAST_X11_TAGS) if Mode.autoinst
+      find_providers(tags)
     end
 
     # List of packages expected to be installed in order to use
@@ -2572,9 +2575,9 @@ module Yast
     #
     # @return [Array<String>] package list
     def remote_x11_packages
-      packages = REMOTE_X11_BASE_PACKAGES.dup
-      packages.concat(AUTOYAST_X11_PACKAGES) if Mode.autoinst
-      packages
+      tags = REMOTE_X11_BASE_TAGS.dup
+      tags.concat(AUTOYAST_X11_TAGS) if Mode.autoinst
+      find_providers(tags)
     end
 
   private
@@ -2624,6 +2627,24 @@ module Yast
           "Failed to select default product pattern %{pattern_name}.\n" +
           "Pattern has not been found."
         ) % {:pattern_name => pattern_name})
+      end
+    end
+
+    # Search for providers for a list of tags
+    #
+    # If a provider is not found, an error will be logged.
+    #
+    # @param tags [Array<String>] List of tags (ie. package names) to search for.
+    # @return [Array<String>] List of packages that provide those tags.
+    def find_providers(tags)
+      tags.each_with_object([]) do |name, providers|
+        provided_by = Pkg.PkgQueryProvides(name).find { |provide| provide[1] != :NONE }
+        if provided_by.nil?
+          providers << name
+          log.error "Provider not found for '#{name}'"
+        else
+          providers << provided_by[0]
+        end
       end
     end
 
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -2658,7 +2658,7 @@ module Yast
       providers = Pkg.PkgQueryProvides(tag).select { |provide| provide[1] != :NONE }
       names = providers.map(&:first)
       if names.size > 1
-        log.info "More than one provider was found for '#{tag}': "\
+        log.warn "More than one provider was found for '#{tag}': "\
           "#{names.sort.join(', ')}."
       end
       names.first
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -2657,11 +2657,12 @@ module Yast
     def find_provider(tag)
       providers = Pkg.PkgQueryProvides(tag).select { |provide| provide[1] != :NONE }
       names = providers.map(&:first)
+      provider = names.include?(tag) ? tag : names.sort.first
       if names.size > 1
         log.warn "More than one provider was found for '#{tag}': "\
-          "#{names.sort.join(', ')}."
+          "#{names.join(', ')}. Selecting '#{provider}'."
       end
-      names.first
+      provider
     end
 
     publish :variable => :install_sources, :type => "boolean"
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -2627,31 +2627,41 @@ module Yast
 
     # Search for providers for a list of tags
     #
-    # The use case of this method is to convert and array
-    # of tags into an array of packages.
+    # The use case of this method is to convert and array of tags into an array
+    # of packages. If a tag does not have a provider, then the tag is included
+    # in the array.
     #
     # @param tags [Array<String>] List of tags (ie. package names) to search for.
     # @return     [Array<String>] List contaning a package for each tag.
     # @see find_provider
     def find_providers(tags)
-      tags.map { |tag| find_provider(tag) }
+      tags.each_with_object([]) do |tag, providers|
+        provider = find_provider(tag)
+        if provider.nil?
+          log.error "Provider not found for '#{tag}'"
+          providers << tag
+        else
+          providers << provider
+        end
+      end
     end
 
     # Search a provider for a tag
     #
     # If a provider is not found, an error will be logged.
     #
-    # @param tag [String] Tag to search a package for.
-    # @return    [String] Name of the package which provides that tag.
+    # @param tag [String]     Tag to search a package for.
+    # @return    [String,nil] Name of the package which provides that tag.
+    #                         It returns nil if no provider is found.
     # @see find_providers
     def find_provider(tag)
-      provided_by = Pkg.PkgQueryProvides(tag).find { |provide| provide[1] != :NONE }
-      if provided_by.nil?
-        log.error "Provider not found for '#{tag}'"
-        tag
-      else
-        provided_by.first
+      providers = Pkg.PkgQueryProvides(tag).select { |provide| provide[1] != :NONE }
+      names = providers.map(&:first)
+      if names.size > 1
+        log.info "More than one provider was found for '#{tag}': "\
+          "#{names.sort.join(', ')}."
       end
+      names.first
     end
 
     publish :variable => :install_sources, :type => "boolean"
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -16,11 +16,9 @@ module Yast
 
     # All known types of resolvables
     RESOLVABLE_TYPES = [:product, :patch, :package, :pattern, :language]
-    # Minimum set of packages required to enable VNC server
+    # Minimum set of packages tags required to enable VNC server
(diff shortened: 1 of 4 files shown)


src/modules/Packages.rb
```diff
@@ -18,6 +18,10 @@ class PackagesClass < Module
     RESOLVABLE_TYPES = [:product, :patch, :package, :pattern, :language]
     # Minimum set of packages required to enable VNC server
     VNC_BASE_PACKAGES = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts", "xinetd"]
+    # Additional packages needed to run second stage in graphical mode
+    # Use exact package name (i.e. libyui-qt6 instead of libyui-qt) to prevent
+    # the UI from asking for confirmation
+    AUTOYAST_X11_PACKAGES = ["libyui-qt6", "yast2-x11"]
```
lslezak: I'd rather avoid `libyui-qt6` as it will be changed at some point, I'd rather use something more stable, like `yast2-qt` (see https://github.com/libyui/libyui-qt/blob/master/libyui-qt.spec.in#L51). @mvidner: or is there any better solution?

ancorgs: As stated in the comment, the problem, according to my experiments, is that using the name of a provide instead of the exact package name, a dialog is displayed to the user with a message "libyui-qt6 is going to be installed". Having extra dialogs is quite bad for autoinstallation. When I use the exact package name the dialog doesn't appear.

Is that the case or am I crazy? In case I'm not crazy, any way to avoid the popup while using a provide?

lslezak: Oh, I overlooked the comment about the popup asking for confirmation... Indeed, that would be bad in autoinstallation.

But maybe we could improve that check to either not ask in AY mode or to use `Provides` instead of the package name...

ancorgs: I originally assumed the popup was being raised somewhere else, in a lower layer. But thinking it twice, it's must be here, in this same module. So I will look for the code raising it and see if I can avoid it in general during autoinstallation (although that's a change in the behavior).

mvidner: Hardcoding the version to libyui-qt6 would be acceptable for a maintenance branch. For master, it will eventually break.
If there is a popup that requires user interaction, that is indeed bad for AY. Usually such popups can be automatically confirmed by using Report.Message instead of Popup.Message in the code. I also wonder whether this happens with all packages that the user has specified via "Provides".

It seems the better solution would be to explicitly resolve the Provides to package names for the packages that we hardcode. But I wasn't able to find the Pkg binding call for that :-/

lslezak: @mvidner there is a [`Pkg.PkgQueryProvides()`](https://github.com/yast/yast-pkg-bindings/blob/master/src/Package.cc#L105) function for resolving provides...

ancorgs: After some extra debugging, here are my findings:

* The dialog is not being shown during the fist attempt of installing the package, but at the end of the installation, during ```Remote#configure_display_manager```
* The first attempt to install the package happens here [1], using ```Pkg::DoProvide```. For "libyui-qt6" it successes. For "libyui-qt" ```Pkg::DoProvide``` returns "the package is not available"

Should we use any alternative to ```DoProvide``` (one supporting provides) on ```SelectSystemPackages``` or should we do that work "by hand" beforehand on ```Packages.vnc_packages```

[1] https://github.com/yast/yast-packager/blob/master/src/modules/Packages.rb#L2161

lslezak: I'd avoid changing the `SelectSystemPackages`, it calls [`SelectSystemPackages`](https://github.com/yast/yast-packager/blob/master/src/modules/Packages.rb#L1137) which is quite complex and changing the simple `DoProvide` based on package names to something else could possibly break something.

I'd rather change `Packages.vnc_packages` to evaluate the provides and return the evaluated package name. This change is better located with smaller impact on the existing code.

imobach:
I've refactored the packages calculation done by `PackagesClass#vnc_packages` and `PackagesClass#remote_x11_packages` so now they are 'tag-based' using the `PkgQueryProvides` method pointed by @lslezak. So, we can specify tags instead of packages.

Also I've changed `libyui-qt6` for `libyui-qt` but I don't know if using `yast2-qt`, as stated by @lslezak, will be a better option.

Thoughts?

src/modules/Packages.rb
```diff
@@ -17,11 +17,15 @@ class PackagesClass < Module
     # All known types of resolvables
     RESOLVABLE_TYPES = [:product, :patch, :package, :pattern, :language]
     # Minimum set of packages required to enable VNC server
-    VNC_BASE_PACKAGES = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts", "xinetd"]
+    VNC_BASE_TAGS = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts", "xinetd"]
+    # Additional packages needed to run second stage in graphical mode
+    # Use exact package name (i.e. libyui-qt6 instead of libyui-qt) to prevent
+    # the UI from asking for confirmation
```
imobach: I guess I should remove this comment.

src/modules/Packages.rb
```diff
@@ -17,11 +17,15 @@ class PackagesClass < Module
...
+    AUTOYAST_X11_TAGS = ["libyui-qt", "yast2-x11"]
     # Default window manager for VNC if none is installed
     DEFAULT_WM = "icewm"
     # Minimum set of packages required for installation with remote X11 server
-    REMOTE_X11_BASE_PACKAGES = [ "xorg-x11-server", "xorg-x11-fonts", "icewm" ]
+    REMOTE_X11_BASE_TAGS = [ "xorg-x11-server", "xorg-x11-fonts", "icewm" ]
```
lslezak: Um, I'm not sure if we should use tags everywhere, what happens if e.g. `icewm` tag is provided by two (or more) packages?

I'd keep this unchanged unless tags are really needed.

src/modules/Packages.rb
```diff
@@ -134,6 +139,8 @@ def main
       @base_source_id = nil
 
       @old_packages_proposal = nil
+
+      PackageSystem.EnsureSourceInit
```
imobach: Is ok to do the initialization here or must I move this call to the `find_providers` method?

src/modules/Packages.rb
```diff
@@ -2555,12 +2562,12 @@ def log_software_selection
...
-      packages = VNC_BASE_PACKAGES.dup
+      tags = VNC_BASE_TAGS.dup
       # At least one windowmanager must be installed (#427044)
       # If none is there, use a fallback
-      packages << DEFAULT_WM unless has_window_manager?
-      packages << "yast2-x11" if Mode.autoinst
-      packages
+      tags << DEFAULT_WM unless has_window_manager?
+      tags.concat(AUTOYAST_X11_TAGS) if Mode.autoinst
+      find_providers(tags)
```
lslezak: `find_providers` returns **all** packages providing the tag. We should choose just one package.

Normally we should get just one package providing the tag, at least for the `libyui-qt` package. But in general there might be more different providers which might even conflict with each other (e.g. `smtp_daemon` is provided by `postfix`, `sendmail`, `exim`, ... and only one of them can be installed).

We should choose only one package providing the tag (and log a warning with the candidates if more packages are found, maybe the user would prefer something else...).

That's why I'd avoid using tags for the other package lists.

src/modules/Packages.rb
lslezak: This initialization is not needed here, it's done outside of the module by the installation framework. You need this only when playing with the package management in a separate script...

lslezak: To be clear: it's not needed at all, just remove it...

src/modules/Packages.rb
```diff
@@ -2623,6 +2630,24 @@ def report_missing_pattern(pattern_name)
...
+    # @param tags [Array<String>] List of tags (ie. package names) to search for.
+    # @return [Array<String>] List of packages that provide those tags.
+    def find_providers(tags)
+      tags.each_with_object([]) do |name, providers|
+        provided_by = Pkg.PkgQueryProvides(name).find { |provide| provide[1] != :NONE }
+        if provided_by.nil?
+          providers << name
+          log.error "Provider not found for '#{name}'"
+        else
+          providers << provided_by[0]
```
lslezak: Prefer `provided_by.first` here.

src/modules/Packages.rb
imobach: Given a package 'blah', maybe we could search first for a package named like that and, if it does not exist, search for a package that provides 'blah'. At this time, we're just taking the first package from `PkgQueryProvides` (I know that you're aware of that; I'm just writing it 'for the record').

src/modules/Packages.rb
```diff
@@ -2623,6 +2625,35 @@ def report_missing_pattern(pattern_name)
...
+    # @param tag [String] Tag to search a package for.
+    # @return    [String] Name of the package which provides that tag.
+    # @see find_providers
+    def find_provider(tag)
+      provided_by = Pkg.PkgQueryProvides(tag).find { |provide| provide[1] != :NONE }
+      if provided_by.nil?
+        log.error "Provider not found for '#{tag}'"
+        tag
+      else
+        provided_by.first
```
lslezak: You should at least log the ignored packages if there are more providers of `tag`. YaST might choose a "random" package if more providers are present. This would help with debugging later...

src/modules/Packages.rb
lslezak: Names first then provides looks as a good solution but IMHO that's over engineering, we should use provides only where needed (i.e. just for the `libyui-qt`), I think it is not needed at the other places...

ancorgs: @lslezak Not sure if I follow you. Do you mean having a separate arrays REMOTE_X11_BASE_TAGS and REMOTE_X11_BASE_PACKAGES? If not, how do you know that checking for provides is needed? On the other hand, checking for the name and then (if not present) for the provides.... doesn't look a bit like reimplementing something that should be resolved by another layer already?

ancorgs: @lslezak I think I got it know. You meant using tags only for AUTOYAST_X11_TAGS and not for the rest. Isn't it? If so, I'm not so sure.

lslezak: @ancorgs Yes, that's what I meant (sorry if that was not clear...). I'm not sure if changing `REMOTE_X11_BASE_PACKAGES` to `REMOTE_X11_BASE_TAGS` is safe...

I mean the package list has never caused any issue (at least I'm not aware of any), changing the behavior might break something. Although it's very unlikely there is a problem with multiple providers which is not handled nicely currently.

Selecting the first found in the list is not reliable (the order is not guaranteed) and this can result in "random" package selection.

I know it's not a problem now, but I know that some AutoYaST users are really creative...

In general I'm not against using the provides, but we have to better handle the corner cases, the current implementation is little bit fragile and non-deterministic ....

src/modules/Packages.rb
```diff
@@ -2623,6 +2625,45 @@ def report_missing_pattern(pattern_name)
...
+    #
+    # @param tag [String]     Tag to search a package for.
+    # @return    [String,nil] Name of the package which provides that tag.
+    #                         It returns nil if no provider is found.
+    # @see find_providers
+    def find_provider(tag)
+      providers = Pkg.PkgQueryProvides(tag).select { |provide| provide[1] != :NONE }
+      names = providers.map(&:first)
+      if names.size > 1
+        log.info "More than one provider was found for '#{tag}': "\
```
lslezak: Yes, this looks better :wink: 

Nit picking: I'd use `log.warn` as this might potentially cause some unexpected behavior...

src/modules/Packages.rb
imobach: Oh, I expected PkgQueryProvides to return a list ordered by some criteria (name, relevance or something like that).

If not, maybe we could select some deterministic heuristic to handle the case of multiple providers. For example, selecting the provider that is called as the tag (if available). If not, just take the first in alphabetical order or something like that. That will be rather easy to implement.

lslezak:
Almost LGTM from my side, we just need to discuss one controversial change:

@mvidner @jreidinger What is your opinion about using provides also for other packages, i.e. changing `REMOTE_X11_BASE_PACKAGES` to `REMOTE_X11_BASE_TAGS`?

Using provides looks as a better solution but there are some drawbacks as I mentioned earlier, I'm not sure if it cannot break something...

(BTW if we go with provides everywhere it should be mentioned in the `.changes` file as this changes behavior for end users...)

jreidinger:
Well, for me using provides make sense, but we need to ensure that we just choose existing one if installed and if not, then I hope solver can choose reasonable one.

mvidner:
AFAIK the API we are changing (packages -> tags) is only used internally. That is, we control the input values. Then IMHO using tags should make our lives easier. If we broke something, we should be covered by tests already existing in openQA, right?

imobach:
@jreidinger: choosing the packages that are installed yet sounds good to me. I'll take care of it.

Thank you all.

imobach:
I've changed the `find_provider` method to prefer an installed provider (if any). Anyway, I'm not sure whether my implementation it's going to work in all cases.

I mean: when a package is `:INST`, what does it mean exactly? I guess that the package is installed and the candidate version (ie. an update) does not satisfy the tag anymore. Or maybe the candidate does not exist. Is this ok?

lslezak:
>  I guess that the package is installed and the candidate version (ie. an update) does not satisfy the tag anymore.

Exactly. See https://github.com/yast/yast-pkg-bindings/blob/master/src/Package.cc#L79

lslezak:
> @jreidinger: choosing the packages that are installed yet sounds good to me. I'll take care of it.

Actually I'm not sure if that's correct, imagine package `A` providing `foo`. At upgrade there is package `B` which provides `foo` and obsoletes `A`. In that case `A` should be upgraded to `B`. Keeping `A` installed sounds wrong to me...

And if dependencies of `A` are not available anymore this might cause a conflict...

(Yeah, working with `Provides` is really tricky...)

imobach:
@lslezak Yes, I read that document but I wasn't sure about implications.

So, `vnc_packages` and `remote_x11_packages` are supposed to return the lists of packages to install. What if I return a package that is it marked as `:INST`? If you try to install that package, eventually you could up trying to update it and the new version won't provide the tag anymore.

lslezak:
I think we should focus on a real use case which is upgrading `libyui-qt6` to `libyui-qt7` (which will definitely happen at some point). How it will behave in that case?

jreidinger:
no other comments then @lslezak

imobach:
@lslezak In that case, I'd expect to have `['libyui-qt6', :INST, :INST], ['libyui-qt7', :CAND, :NONE]` from `PkgQueryProvides`. So I think that considering only :CAND and :BOTH alternatives may be safe.

But I'm just guessing. I need to do more research about that.

lslezak:
Ok, so then the code should prefer `:BOTH` (the same package as installed, just updated) then `:CAND` (new package, probably after rename).

imobach:
@lslezak Yes, it sounds like a good approach. My only doubt is what to do with packages that are in an `:INST` stage (and with no `:CAND` alternatives). We could:

* Just ignore them (logging and error and removing them from the list).
* Add them to the list (and maybe logging this situation).

Thoughs?

lslezak:
I think we can simply ignore that case, normally that should not happen (removing `libyui-qt` from installation repositories is really unlikely)...

Just logging an error when no package providing the tag is found is sufficient (that's already present in the PR), that case could indicate a repository setup error.

imobach:
Ok, I've submitted a new commit to implement that behaviour. `:BOTH` and `:CAND` are preferred now. I guess it should handle installation and also update.

jreidinger:
LGTM

lslezak:
LGTM, thanks a lot! :+1:
It turned out to be quite complicated... :sweat_smile:

imobach:
Thank you all for your patience :)

PR #123, opened 2015-06-02
Fixed building a local ISO URL (bnc#919138)

lslezak:
The `url` parameter for a local ISO URL needs to use `dir` scheme, not `iso`.

E.g. the final URL should be `iso:///?iso=openSUSE-13.2-DVD-x86_64.iso&url=dir%3A%2Finstall`, not `iso:///?iso=openSUSE-13.2-DVD-x86_64.iso&url=iso%3A%2Finstall`

- 3.1.69

Fixes https://bugzilla.suse.com/show_bug.cgi?id=919138.

### DIFF
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -333,6 +333,7 @@ module Yast
       new_url = uri.dup
       new_url.path = File.dirname(uri.path || "")
       new_url.query = nil
+      new_url.scheme = "dir" if uri.scheme.downcase == "iso"
       params["url"] = new_url.to_s
 
       processed = URI("")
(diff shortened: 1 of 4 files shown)


lslezak:
One of the Travis builds fails, it cannot checkout the commit (huh, ??), but we can ignore that. The second (more important) merged build passed.

src/modules/SourceDialogs.rb
```diff
@@ -333,6 +333,7 @@ def PostprocessISOURL(url)
       new_url = uri.dup
       new_url.path = File.dirname(uri.path || "")
       new_url.query = nil
+      new_url.scheme = "dir" if uri.scheme.downcase == "iso"
```
jreidinger: please add comment with bnc, so we know in future why it is needed.

jreidinger:
thanks, LGTM

PR #124, opened 2015-06-05
use the full URL when editing a repository URL (bnc#826075)

lslezak:
the full URL includes also the password (if set)

- 3.1.70


jreidinger:
LGTM

mvidner:
The surrounding method and file are huge. We should find an opportunity to improve: coverage report, refactor, test. https://codeclimate.com/github/yast/yast-packager

PR #126, opened 2015-06-17
Enabled YARD docs for rubydoc.info

mvidner:
In https://github.com/yast/yast-kdump/pull/30 we use an API from this package, so it should have documentation.


kobliha:
OK, looks good

PR #127, opened 2015-06-22
Recommended pkgs: Do not offer to install for already installed ones on the first run

mvidner:
https://bugzilla.opensuse.org/show_bug.cgi?id=902394 , part of https://fate.suse.com/318099

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jun 19 13:46:42 UTC 2015 - mvidner@suse.com
+
+- Do not offer to install recommended packages for already installed
+  ones on the first run (boo#902394).
+- 3.1.72
+
+-------------------------------------------------------------------
 Thu Jun 11 14:58:58 UTC 2015 - lslezak@suse.cz
 
 - proper fix for a float remaining time (bnc#882240)
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        3.1.71
+Version:        3.1.72
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
--- a/src/clients/sw_single.rb
+++ b/src/clients/sw_single.rb
@@ -594,11 +594,6 @@ module Yast
             opts = GetPackagerOptions()
             Builtins.y2milestone("Using packager widget options: %1", opts)
 
-            home = ENV["home"] || "/root"
-            # unset the ignoreAlreadyRecommended flag on the very first run,
-            # otherwise keep the value from the config file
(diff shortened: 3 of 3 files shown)


kobliha:
This PR removes exactly this commit https://github.com/yast/yast-packager/commit/e634cb127afd2a7fd05a297bf81eed615011a9cf

kobliha:
... but it seems to be what we want as the NEW implementation will give users the choice to [x] Always install also recommended ... or not.

So, LGTM

mvidner:
Ah, thanks, that one links https://bugzilla.opensuse.org/show_bug.cgi?id=835968 which is another report of the "Live CD" use case. I will need to check the code locations mentioned in  https://bugzilla.novell.com/show_bug.cgi?id=811237 which is linked from the 1st bug.

gabi2:
To my understanding and what Michel (ma) explained to me we shouldn't set/reset the solver ignoreAlreadyrecommededOption in YaST but let it set to libzypp default. It will be set accordingly, i.e. as soon as solver implementation is fixed to respect only language/hardware/filesystem dependencies for already installed packages it will be set 'false'.

mvidner:
Hmm, the flag is set in one more place in this package, https://github.com/yast/yast-packager/blob/5b3a37da3f62fed4f9e6e296efc1785db9491288/src/modules/Packages.rb#L2222
Packages.Proposal:
```ruby
 Pkg.SetSolverFlags({ "ignoreAlreadyRecommended" => Mode.normal })
```
So I won't rush this.

gabi2:
Another place... flag set to Mode.normal?It gets more and more complicated.

gabi2:
@kobliha The new option 'Install Recommeded Packages' applies to "normal" recommends whereas the option 'IgnoreRecommendedForAlreadyInstalledPackages' / the solver flag 'ignoreAlreadyRecommended' concerns dependencies of already installed packages.

gabi2:
@mvidner I have talked again to Michel (ma). 
My conclusion that libzypp default for "ignoreAlreadyRecommended"will be set 'false' (as soon as solver is fixed) was wrong. Michel explained that default will remain 'true'; to his opinion there shouldn't be additional packages installed except the user knows about them and explicetly wants it that way.
With removing the option from UIs we wouldn't have any possibility to install them, e.g. if a new language is selected trigger installation of language packages for already installed packages/applications. A solution might be to offer the option again in menu but do not store it.
A good solution for languages can be to offer the possibility to install all language packages for already installed packages (set the solver flag for this run) in a popup when selecting a language for installation in 'Filter' Languages.

gabi2:
Concerning the Live-CD: 
this use case indeed requires setting "ignoreAlreadyRecommended" to 'false'.

mvidner:
> ```ruby
Pkg.SetSolverFlags({ "ignoreAlreadyRecommended" => Mode.normal })
```

That is to support the non-package triggers (hardware, or in this case btrfs -> btrfsprogs), see https://bugzilla.suse.com/show_bug.cgi?id=668588 , so I'll leave that in.

gabi2:
LGTM

PR #131, opened 2015-07-21
Fix dvd

jreidinger:


package/yast2-packager.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 21 08:32:41 UTC 2015 - jreidinger@suse.com
+
+- fix crash caused by path t DVD that does not contain path
```
ancorgs: t -> to. Moreover, not the most understandable entry ever. :wink:

ancorgs:
Good enough for a quick fix. Both code and test would benefit from some cleanup, but first things first.

PR #132, opened 2015-08-28
Do not show a warning for the removed SMT product (bsc#942639)

lslezak:
...when upgrading from SLES11 + SMT

(the SMT functionality has been integrated into SLES12)

- 3.1.77

Details: there is an internal mapping with known product renames, when a replacement is found in the mapping the product removal warning is not displayed.


schubi2:
LGTM

PR #134, opened 2015-10-06
Repos without service

jreidinger:


src/clients/repositories.rb
```diff
@@ -171,6 +171,7 @@ def PriorityToString(priority)
 
     def ReposFromService(service, input)
       input = deep_copy(input)
+      service = "" if service == :without
```
lslezak: `:without` is used at several places, we should define a constant for it.

(And `:without` symbol sounds a bit strange for me, I'd rather prefer `:no_service` or something like that.)

src/clients/repositories.rb
```diff
@@ -641,10 +642,20 @@ def buildList
...
         )
         ret = Builtins.add(ret, t)
-      end 
+      end
 
+      # there is some service, so allow to filter repos without service (bnc#944504)
+      if ret.size > 2
+        t = Item(
+          Id(:without_service),
+          _("Without Service"),
```
lslezak: It's the other way round, a service contains repositories. The filter should be named "Repositories not in a service",  "Repositories not belonging to a service" or something like this. A bit longer but I'm not sure if there is a good shorter alternative...

@mvidner any suggestion here?

package/yast2-packager.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Oct  6 11:13:46 UTC 2015 - jreidinger@suse.com
+
+- If there is service allow to see repositories without service
```
mvidner: > If there are services, allow to see repositories not belonging to a service

mvidner: Ah, only now I understood the bug report.
> Added a filter showing only those repositories not belonging to a service.

src/clients/repositories.rb
mvidner: "Only repositories not provided by a service"

kobliha: And then definitely add # TRANSLATORS: Explanation what it actually means :) and where it is used

src/clients/repositories.rb
```diff
@@ -641,10 +642,20 @@ def buildList
             @displayed_service == Ops.get_string(srv_state, "alias", "")
         )
         ret = Builtins.add(ret, t)
-      end 
+      end
 
+      # there is some service, so allow to filter repos without service (bnc#944504)
+      if ret.size > 2
```
mvidner: `size > 2` is very non-obvious. how about `unless Builtins.isempty(@serviceStatesOut)`

src/clients/repositories.rb
jreidinger: OK, will add TRANSLATORS, just improved a bit mvidner text with replaced "a" with "any" as there is usually more services there, so not provided by any service.

src/clients/repositories.rb
jreidinger: will fix

src/clients/repositories.rb
jreidinger: that is reason why I add comment there. Basically I am not fan of adding new buildins and without it we need to be very care to not cause exception

lslezak:
LGTM, do not forget to send the new POT file to Karl, see the "Text Freeze, SP1" mail on yast-devel@.

PR #136, opened 2015-10-08
repositories - fixed switch between dvd:// and cd:// URL (bsc#947595)

lslezak:
Libzypp failed with `Url scheme requires path name` for `cd://` or `dvd://` URL, it requires `cd:///` and `dvd:///`.

Tested manually.


jreidinger:
LGTM

PR #137, opened 2015-10-16
repository manager - properly close the main window (bsc#950525)

lslezak:
...when aborting after failed initialization to avoid a segfault

See [bsc#950525](https://bugzilla.suse.com/show_bug.cgi?id=950525)

- 3.1.82


mchf:
LGTM

PR #138, opened 2015-10-21
Product license nil check

lslezak:
This is a fix for a P1 bug [bsc#951210](https://bugzilla.suse.com/show_bug.cgi?id=951210).

- Only the firstboot sequence is affected, in a standard installation/upgrade the `src_id` is defined (it refers to the installation repository) and is not `nil`.

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -1087,7 +1087,7 @@ module Yast
 
 
     # Ask user to confirm license agreement
-    # @param [Fixnum] src_id integer repository to get the license from.
+    # @param [Fixnum,nil] src_id integer repository to get the license from.
     #   If set to 'nil', the license is considered to belong to a base product
     # @param [String] dir string directory to look for the license in if src_id is nil
     #   and not 1st stage installation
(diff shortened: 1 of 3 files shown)


imobach:
LGTM

src/modules/ProductLicense.rb
```diff
@@ -1145,12 +1145,13 @@ def AskLicenseAgreement(src_id, dir, patterns, action, enable_back, base_product
...
       licenses_ref = arg_ref(licenses)
 
-      label = Pkg::SourceGeneralData(src_id)["name"]
       title = _("License Agreement")
 
-      if !label.empty?
-        # %s is an extension name, e.g. "SUSE Linux Enterprise Software Development Kit"
-        title = _("%s License Agreement") % label
+      if src_id
+        label = Pkg::SourceGeneralData(src_id)["name"]
```
kobliha: Can `Pkg::SourceGeneralData(src_id)` return nil?

src/modules/ProductLicense.rb
```diff
@@ -1145,12 +1145,13 @@ def AskLicenseAgreement(src_id, dir, patterns, action, enable_back, base_product
...
+        # TRANSLATORS: %s is an extension name
+        # e.g. "SUSE Linux Enterprise Software Development Kit"
```
kobliha: Are there any limitations for length? As short as possible? Should be in the comment

lslezak: The text to translate is actually short, it's just `%s License Agreement`, no need for comments.

The longest part is the product name anyway. But even a long product name is OK, it's used a title which is split to multiple lines if needed.

src/modules/ProductLicense.rb
lslezak: Actually yes, if you use a wrong `src_id`...

src/modules/ProductLicense.rb
kobliha: OK, then it doesn't need the comment :)

kobliha:
LGTM

lslezak:
Merging, got the fix confirmation from the reporter.

PR #140, opened 2015-11-11
Do not check the free space on a CD/DVD mounted medium (bsc#952112)

lslezak:
... during online migration

the mount point has been changed from `/media` to `/run/media`


ancorgs:
LGTM... Although we could probably keep both checks, (/media || /run/media). You never know when they will change mind again. :-)

ancorgs:
LGTM

lslezak:
@ancorgs Thanks for the review!

PR #142, opened 2015-12-16
Properly abort the addon workflow (bsc#956597)

lslezak:
... when adding a new addon fails, avoid removing a system repo by mistake.

- 3.1.84

Notes: The code at [this line](https://github.com/yast/yast-packager/blob/c3ff51a53b527b171ac701c3f6d1f9d89f125301/src/include/packager/repositories_include.rb#L349) returns `:cancel` which was treated as a success.

The fix has been tested in SLE12-SP1-GM, the exact steps how to reproduce the problem are described at [this bugzilla comment](https://bugzilla.suse.com/show_bug.cgi?id=956597#c4).

### DIFF
--- a/src/include/packager/repositories_include.rb
+++ b/src/include/packager/repositories_include.rb
@@ -464,7 +464,7 @@ module Yast
 
       if ret == :again
         return :back
-      elsif ret == :abort
+      elsif ret == :abort || ret == :cancel
         return :abort
       end
       :next
--- a/src/include/packager/repositories_include.rb
+++ b/src/include/packager/repositories_include.rb
@@ -462,12 +462,17 @@ module Yast
 
       ret = createSource(url, plaindir, @download_meta, name)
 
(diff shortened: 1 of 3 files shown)


src/include/packager/repositories_include.rb
```diff
@@ -464,7 +464,7 @@ def StoreSource
 
       if ret == :again
         return :back
-      elsif ret == :abort
+      elsif ret == :abort || ret == :cancel
         return :abort
       end
       :next
```
jreidinger: what about changing this whole paragraph into
```
case ret
when :again 
  :back
when :abort, :cancel
  :abort
else
  :next
end
```

lslezak: OK, makes sense, originally I wanted as minimal change as possible, but this looks really better...

jreidinger:
LGTM

PR #143, opened 2015-12-21
Remove leftover richtext tag

lslezak:
Fixes https://bugzilla.suse.com/show_bug.cgi?id=950866

(`master` only to not break the existing translations.)


imobach:
LGTM

PR #146, opened 2016-01-27
Do not select additional products when initializing the default packages

lslezak:
The code resets the package selection and re-selects the default products. Unfortunately it did not check the original product status and selected all available products even those originally not selected to install.

The problem happens only with the SLE-for-SAP product because during registration it receives also the HA repository and the standard SLES repo. And the products from these additional repositories gets selected by mistake.

### The Fix

Originally also the standard SLES and HA products were preselected by default:

sles-sap-broken-products

After applying the fix only the SLES-for-SAP product is selected:

sles-sap-fixed-products

### Notes

- Added test cases
- Rubyfied the code a bit

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Jan 27 12:16:51 UTC 2016 - lslezak@suse.cz
+
+- Fixed selecting additional products in the SLES-for-SAP product
+  (do not select previously unselected products when initializing
+  the default package selection) (bsc#963036)
+- 3.1.84.1
+
+-------------------------------------------------------------------
 Mon Jan  4 09:33:33 UTC 2016 - jreidinger@suse.com
 
 -  do not crash if adding unknown repo type (bnc#960460) 
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,9 +1,9 @@
 -------------------------------------------------------------------
 Wed Jan 27 12:16:51 UTC 2016 - lslezak@suse.cz
 
-- Fixed selecting additional products in the SLES-for-SAP product
-  (do not select previously unselected products when initializing
-  the default package selection) (bsc#963036)
+- Fixed selecting additional products during system installation
+  (do not select previously unselected products after adding
+  repositories from the registration server) (bsc#963036)
 - 3.1.84.1
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


test/packages_test.rb
```diff
@@ -836,4 +847,32 @@ def pattern(properties = {})
       end
     end
   end
+
+  describe "#Reset" do
```
kobliha: Just a nitpicking: I know that a test should be self-describing, but could you add more comments to this behavior? Those `it` parts are somehow `not enough` understandable to me. You could maybe define some `when` clauses.

package/yast2-packager.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Jan 27 12:16:51 UTC 2016 - lslezak@suse.cz
+
+- Fixed selecting additional products in the SLES-for-SAP product
```
kobliha: Nitpicking: the product itself is not important, the fixed/changed behavior seems to be more important than that. I might not even mention SAP here.

kobliha:
I did not check the code, I've checked this PR only from the nitpicking POV. Sorry.

test/packages_test.rb
```diff
@@ -836,4 +847,39 @@ def pattern(properties = {})
...
+    # Reset package changes done by YaST but select the installe products back.
```
kobliha: installe -> d

test/packages_test.rb
```diff
@@ -836,4 +847,39 @@ def pattern(properties = {})
...
+    # Do not select previously not selected products which were added by
```
kobliha: Sounds quite cryptic :)

test/packages_test.rb
```diff
@@ -836,4 +847,39 @@ def pattern(properties = {})
...
+        [product("name" => "p1", "status" => :selected), product("name" => "p2")]
+      )
+
+      expect(Yast::Pkg).to receive(:ResolvableInstall).with("p1", :product)
+      expect(Yast::Pkg).not_to receive(:ResolvableInstall).with("p2", :product)
+
+      Yast::Packages.Reset([:product])
+    end
+
+    # When restoring the selected products after reset ignore the products
```
kobliha: Captial R in Reset?

missing a comma after reset

the products -> all products

kobliha:
Comments in test are definitely better than nothing, but as I said, I did not check the code itself

kobliha:
LGTM for comments :)

PR #147, opened 2016-01-29
Merge the SP1 product selection fix to master

lslezak:


src/modules/Packages.rb
```diff
@@ -788,30 +788,29 @@ def ForceFullRepropose
...
-      Builtins.foreach(keep) do |type|
-        selected = Pkg.ResolvableProperties("", type, "")
-        Builtins.foreach(selected) do |s|
-          restore = Builtins.add(
-            restore,
-            { "type" => type, "name" => Ops.get_string(s, "name", "") }
-          )
+
+      # collect the currently selected resolvables
+      keep.each do |type|
```
jreidinger: what about using `each_with_object`? It basically allows also quick skip like
```
restore = resolvables.each_with_object([]) do |resolvable, restore|
  next if resolvable["status"] != :selected || resolvable["transact_by"] == :solver

  restore << { "type" => type, "name" => resolvable["name"] }
end
```

src/modules/Packages.rb
```diff
@@ -788,30 +788,29 @@ def ForceFullRepropose
...
+      # This keeps the user-made changes (BNC#446406)
       Pkg.PkgApplReset
-      Builtins.foreach(restore) do |res|
-        Pkg.ResolvableInstall(
-          Ops.get_string(res, "name", ""),
-          Ops.get_symbol(res, "type")
-        )
-      end
+
+      restore.each { |res| Pkg.ResolvableInstall(res["name"], res["type"]) }
```
jreidinger: when I see usage of map, I think it is a bit overkill to use hash here. simple pair is enough. and you can name it.
so above you can add `restore << [ type, resolvable["name"]]`
and here `restore.each { |type, name| Pkg.ResolvableInstall(name, type) }`

test/packages_test.rb
```diff
@@ -836,4 +847,36 @@ def pattern(properties = {})
...
+
+      allow(Yast::Pkg).to receive(:ResolvableProperties).and_return(
+        [product("name" => "p1", "status" => :selected, "transact_by" => :solver)]
+      )
+
+      expect(Yast::Pkg).not_to receive(:ResolvableInstall).with("p1", :product)
+
+      Yast::Packages.Reset([:product])
+    end
+  end
```
jreidinger: nice to have test :+1:

src/modules/Packages.rb
lslezak: But we need to run this loop for each `keep` value, we would need to merge the results....

```ruby
restore.merge(resolvables.each_object ....)
```

I'm not sure if that would better readable than the current code... :worried:

jreidinger: actually not, as you can pass it with:
```
restore = keep.each_with_object([]) do |type, restore|
  resolvables.each_with_object(restore) do |resolvable, restore|
```

jreidinger:
just NP, so LGTM

lslezak:
OK, thanks!

PR #149, opened 2016-02-23
Handle update messages

imobach:
This PR uses the feature added to yast-yast2 in https://github.com/yast/yast-yast2/pull/442. It shows a popup showing packages notifications during normal and upgrade mode. This message is shown, for example, in `sw_single` or during a regular system upgrade.

### DIFF
--- a/test/package_installation_test.rb
+++ b/test/package_installation_test.rb
@@ -18,15 +18,38 @@ describe Yast::PackageInstallation do
       allow(Yast::PackagesUI).to receive(:show_update_messages)
     end
 
-    context "when running in normal mode" do
-      it "shows a summary" do
-        expect(Yast::PackagesUI).to receive(:SetPackageSummary)
-        subject.Commit(config)
-      end
-
+    context "when commit is successful" do
       it "returns the commit result" do
         allow(Yast::PackagesUI).to receive(:SetPackageSummary)
         expect(subject.Commit(config)).to eq(result)
+      end
+
+      context "when running in normal mode" do
+        it "shows a summary" do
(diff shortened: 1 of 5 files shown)


test/package_installation_test.rb
```diff
@@ -0,0 +1,84 @@
...
+          "currentPath"      => "/var/adm/update-message/dummy-package-1.0",
+        }
+      end
+
+      it "shows the update messages" do
+        expect(Yast::PackagesUI).to receive(:show_update_messages).with(result)
+        subject.Commit(config)
+      end
+
+      context "in installation mode" do
```
lslezak: NP: the contexts are a bit mixed IMHO:

- normal mode
- update messages received
  - installation mode
- fails

I'd rather expect this structure:

- normal mode
  - fails
- installation mode
  - update messages received
  - fails

(You can share the tests for the "fails" part.)

Or alternatively

- fails
  - normal mode
  - installation mode
- update messages received
  - normal mode
  - installation mode

I'm not sure which one is better but I'd rather avoid mixing the contexts if possible...

lslezak:
One test fails at Travis.

imobach:
You're right (I written those tests too late at night :) ). I've reorganized them:

```
when commit is successful
  returns the commit result
  when running in normal mode
    shows a summary
  when not running in normal mod
    does not show a summary
when installation fails
  logs the error and returns []
when update messages are received
  shows the update messages
  in installation mode
    does not show the update messages
  in autoinstallation mode
    does not show the update messages
```

lslezak:
OK, LGTM now :+1:

imobach:
Thank you!

PR #151, opened 2016-04-01
Added Network Configuration button to Source Dialogs (Fate#319716)

teclator:
The button added should be available only during installation as in a running system you should have network configured and doesn't make sense in auto_installation

### DIFF
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -87,6 +87,7 @@ module Yast
       Yast.import "ProductControl"
       Yast.import "ProductFeatures"
       Yast.import "Stage"
+      Yast.import "WFM"
 
       # common functions / data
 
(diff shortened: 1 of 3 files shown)


ancorgs:
LGTM

src/modules/SourceDialogs.rb
```diff
@@ -2013,6 +2015,14 @@ def scc_repos_widget
       display_scc ? Left(RadioButton(Id(:sccrepos), _(WIDGET_LABELS[:sccrepos]))) : Empty()
     end
 
+    def network_button
+      if Mode.installation || Mode.live_installation || Mode.update
```
kobliha: JFYI, `Mode.installation` is `true` for modes: installation, autoinstallation, live_installation

teclator: You are right, as same update is true for "update" and "autoupgrade"

PR #152, opened 2016-04-04
using uptime

schubi2:


lslezak:
Code wise LGTM, please check why the `package_installation_test.rb` test fails at Travis.

schubi2:
I have built it directly in YaST:Devel without any problems. So, I am merging.....

PR #154, opened 2016-04-11
Select the default product patterns

lslezak:
... defined by the `defaultpattern()` provides.

See [FATE#320199](https://fate.suse.com/320199).

Manually tested in a patched installer.

### DIFF
--- a/src/lib/packager/product_patterns.rb
+++ b/src/lib/packager/product_patterns.rb
@@ -50,7 +50,7 @@ module Yast
     # @return [Array<String>] pattern names
     def find
       products = Yast::Pkg.ResolvableProperties("", :product, "")
-      products.select! { |product| product["status"] == :selected }
+      remove_unselected(products)
       products.map! { |product| product["name"] }
       log.info "Found selected products: #{products}"
 
@@ -78,15 +78,14 @@ module Yast
       product_dependencies = []
 
       resolvables = Yast::Pkg.ResolvableProperties(product, :product, "")
-      resolvables.select! { |p| p["status"] == :selected }
+      remove_unselected(resolvables)
 
       resolvables.each do |resolvable|
         prod_pkg = resolvable["product_package"]
         next unless prod_pkg
 
         release_resolvables = Yast::Pkg.ResolvableDependencies(prod_pkg, :package, "")
-        next unless release_resolvables
-        release_resolvables.select! { |p| p["status"] == :selected }
+        remove_unselected(release_resolvables)
 
         release_resolvables.each do |release_resolvable|
           deps = release_resolvable["deps"]
@@ -97,6 +96,12 @@ module Yast
       log.debug "Product #{product} depependencies: #{product_dependencies}"
 
       product_dependencies
+    end
+    # Remove not selected resolvables from the list
+    # @param [Array<Hash>] resolvables only the Hashes where the key "status"
+    #   maps to :selected value are kept, the rest is removed
+    def remove_unselected(resolvables)
+      resolvables.select! { |p| p["status"] == :selected }
     end
 
     # Collect "provides" dependencies from the list.
--- a/src/lib/packager/product_patterns.rb
+++ b/src/lib/packager/product_patterns.rb
@@ -1,0 +1,141 @@
+
+# ------------------------------------------------------------------------------
+# Copyright (c) 2016 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+#
+
+require "yast"
+
+module Packager
+
+  # Evaluate the default patterns for the currently selected products
+  class ProductPatterns
+    include Yast::Logger
+
+    def initialize
+      Yast.import "Pkg"
+    end
+
+    # Find the default patterns for all selected products.
+    # @note The package management and the products needs to be initialized and
+    #   selected *before* using this method.
+    # @return [Array<String>] pattern names
+    def names
+      @names ||= find
+    end
+
+    # Select the default patterns to installation
+    # @note The package management and the products needs to be initialized and
+    #   selected *before* using this method.
+    # @return [Boolean] true if all patterns were successfuly selected,
+    #   false otherwise
+    def select
+      # use map + all? to try installing *all* patterns, plain all? would stop
+      # at the first failure
+      names.map { |p| Yast::Pkg.ResolvableInstall(p, :pattern) }.all?
+    end
+
+  private
+
+    # Find the default patterns for all selected products.
+    # @return [Array<String>] pattern names
+    def find
+      products = Yast::Pkg.ResolvableProperties("", :product, "")
+      products.select! { |product| product["status"] == :selected }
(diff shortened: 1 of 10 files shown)


src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,141 @@
...
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+#
+# ------------------------------------------------------------------------------
+#
+
+require "yast"
+
+module Packager
```
mchf: just a question ... why is it not encapsulated in ```Yast``` module too?

src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+  private
+
+    # Find the default patterns for all selected products.
+    # @return [Array<String>] pattern names
+    def find
+      products = Yast::Pkg.ResolvableProperties("", :product, "")
+      products.select! { |product| product["status"] == :selected }
+      products.map! { |product| product["name"] }
```
mchf: Is it possible that one product is in new array multiple times or not?

src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,141 @@
...
+      resolvables = Yast::Pkg.ResolvableProperties(product, :product, "")
+      resolvables.select! { |p| p["status"] == :selected }
+
+      resolvables.each do |resolvable|
+        prod_pkg = resolvable["product_package"]
+        next unless prod_pkg
+
+        release_resolvables = Yast::Pkg.ResolvableDependencies(prod_pkg, :package, "")
+        next unless release_resolvables
+        release_resolvables.select! { |p| p["status"] == :selected }
```
mchf: this is very similar to the selecting ```:selected``` in ```find```. Does it make sense to put it into own method?

src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      resolvables = Yast::Pkg.ResolvableProperties(product, :product, "")
+      resolvables.select! { |p| p["status"] == :selected }
+
+      resolvables.each do |resolvable|
+        prod_pkg = resolvable["product_package"]
+        next unless prod_pkg
+
+        release_resolvables = Yast::Pkg.ResolvableDependencies(prod_pkg, :package, "")
+        next unless release_resolvables
```
mchf: can ```release_resolvables``` be nil here? If yes then ```find``` code has potential exception inside (line 53). If no then the ```next``` is useless.

src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,141 @@
...
+    # @param [Array<Hash>] dependencies all dependencies
+    # @return [Array<String>] only the "provides" dependencies
+    # @example
+    #   provides([{"provides" => "foo"}, {"requires" => "bar"}, ...]) => ["foo"]
+    def provides(dependencies)
+      provides = []
+
+      dependencies.each do |dependency|
+        prov = dependency["provides"]
+        provides << prov if prov
```
mchf: what if ```"provides"``` is not int the hash? Can it happen?

src/lib/packager/product_patterns.rb
lslezak: There can be several versions of the same product in the list but only one of them can be in `:selected` state. There should be no duplicates.

src/lib/packager/product_patterns.rb
lslezak: Is that always needed? E.g. bootloader uses [Bootloader](https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/device_map_dialog.rb#L10), nfs-client [Fstab::TSort](https://github.com/yast/yast-nfs-client/blob/master/src/lib/fstab/tsort.rb#L4), IIRC the convention is to use the package name.

@jreidinger ?

mchf: I don't know that's why I'm asking ... I've just thought that it is our "playground"

src/lib/packager/product_patterns.rb
lslezak: Yes, `provides` might be missing. In that case it's ignored by the `if prov` condition.

src/lib/packager/product_patterns.rb
jreidinger: If you do not use Yast namespace, then there is no risk of collision. On other hand, if you use Yast namespace, you do not need to prefix all calls to Yast with namespace. So main question is how closely you want to have your code tighten to Yast itself. So it have pros and cons and it is up to you to decide.

mchf: I would mainly ask about convention. Above pros / cons are more less clear. Is ```Yast``` namespace intended as namespace for all our code or for old converted code only or ... ?

src/lib/packager/product_patterns.rb
mchf: ok, thanks

src/lib/packager/product_patterns.rb
mchf: ok

src/lib/packager/product_patterns.rb
jreidinger: Well, convention is not set in stone. My personal convention is that something YaST specific go to Yast namespace and something that is more independent is more likely should go to own namespace, so we can in future separate it if needed.

mvidner: IMO ::Yast is the legacy namespace converted from YCP. That is AFAIK what the existing `/usr/share/YaST2/lib/` libraries conform to.

lslezak: Agreed with @jreidinger. As this code is quite tight to YaST then moving to `Yast` name space make sense here.

mchf: that's what i wanted to know ;-)

src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,146 @@
...
+
+    def initialize
+      Yast.import "Pkg"
+    end
+
+    # Find the default patterns for all selected products.
+    # @note The package management and the products needs to be initialized and
+    #   selected *before* using this method.
+    # @return [Array<String>] pattern names
+    def names
```
mchf: as this is part of public API, I'm thinking if there is proper place for raising an exception when package management is not initialized as expected. Is it possible to detect it somehow?

src/lib/packager/product_patterns.rb
```diff
@@ -0,0 +1,146 @@
...
+        release_resolvables.each do |release_resolvable|
+          deps = release_resolvable["deps"]
+          product_dependencies.concat(deps) if deps
+        end
+      end
+
+      log.debug "Product #{product} depependencies: #{product_dependencies}"
+
+      product_dependencies
+    end
```
mchf: NP: empty line

src/lib/packager/product_patterns.rb
lslezak: Actually I'm not sure if that would be possible, maybe it would be a bit tricky... Anyway it's not only about the initialization, also the correct products need to selected to install. So I'd leave the responsibility on the caller to properly set all what's needed.

src/lib/packager/product_patterns.rb
lslezak: I have reformatted the files with Rubocop, it found several issues, but it did not complain at this place. Let's consider is as OK :wink:

src/lib/packager/product_patterns.rb
mchf: ok

mchf:
LGTM

PR #155, opened 2016-04-22
Exposed methods to mark info as seen in Installation (bsc#974409)

teclator:
It is related to yast/yast-installation#367


ancorgs:
LGTM

PR #156, opened 2016-04-22
Move the default pattern selection to a different place (FATE#320199)

lslezak:
...to not conflict with the pattern selection in the system upgrade mode

`ComputeSystemPatternList` is called also in the upgrade mode, but in upgrade mode we need to select the default product patterns only in the *pattern based* upgrade.

In the *packages based* upgrade the default patterns must not be selected to avoid adding new packages into the system.

The `patterns_to_install` method is called only in the installation mode.

- 3.1.96


ancorgs:
lgtm

PR #157, opened 2016-05-10
Disable local repositories

imobach:
After installing from a local media, like a CD/DVD or a USB stick, the installation source is still registered as a repository. Sadly it could cause some issues installing or updating packages.

Fortunately, this behavior will change in future SUSE/openSUSE versions. Now, at the end of the installation, all local repositories will be checked (including the installation media) and:

* if they contain products that are available through remote sources, they will be disabled;
* if the product is not available by other means, it will be left untouched.

# Not to blog

* Refactor (and test) `PkgFinishClient` to inherit from `Installation::FinishClient`.
* Modify that client to disable uneeded repositories (CD/DVD/ repositories whose products are available through other repositories).
* Depends on https://github.com/yast/yast-yast2/pull/465


src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,171 @@
...
+
+module Yast
+  class PkgFinishClient < ::Installation::FinishClient
+    include Yast::I18n
+    include Yast::Logger
+
+    # Path to libzypp repositories
+    REPOS_DIR = "/etc/zypp/repos.d"
+    # Repository schemes to disable (disable_local_repos)
+    SCHEMES_TO_DISABLE = [:cd, :dvd]
```
lslezak: [Feature#320494](https://fate.suse.com/320494) also mentions an USB flash disk. That would be the `hd` schema.

BTW libzypp considers the "cd", "dvd", "dir", "hd", "iso", "file" schema as local, see https://github.com/openSUSE/libzypp/blob/a7a038aeda1ad6d9e441e7d3755612aa83320dce/zypp/Url.cc#L458

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,171 @@
...
+      # save all repositories and finish target
+      Pkg.SourceSaveAll
+      Pkg.TargetFinish
+
+      # save repository metadata cache to the installed system
+      # (needs to be done _after_ saving repositories, see bnc#700881)
+      Pkg.SourceCacheCopyTo(Installation.destdir)
+
+      # copy list of failed packages to installed system
+      WFM.Execute(
```
lslezak: NTH: I'd prefer refactoring this and using the ``::File.exist?` and `::FileUtils.cp_a` from Ruby stdlib here.

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,171 @@
...
+      current_repos = SCR.Read(path(".target.dir"), REPOS_DIR)
+
+      if current_repos.nil? || current_repos.empty?
+        log.warn("There are currently no repos in #{REPOS_DIR} conf dir")
+        return
+      else
+        log.info("These repos currently exist on a target: #{current_repos}")
+      end
+
+      cmd = WFM.Execute(path(".local.bash_output"), "date +%Y%m%d-%H%M%S")
```
lslezak: [Time.strftime](http://ruby-doc.org/core-2.2.0/Time.html#method-i-strftime) would be nice here, but it seems that lot of places in this client could be Rubyfied... So maybe at least few of them could be improved... :wink:

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,171 @@
...
+    # * if some product contained is not available through another
+    #   repository, then it will be left untouched.
+    #
+    # @param schemes [Array<Symbol>] Schemes to consider
+    # @return [Array<Packages::Repository>] List of disabled repositories
+    def disable_uneeded_repos(schemes)
+      candidates_repos, other_repos = *::Packages::Repository.enabled.partition do |repo|
+        schemes.include?(repo.scheme)
+      end
+      product_names = other_repos.map(&:products).flatten.map(&:name)
```
lslezak: So we just evaluate the plain product names? I still think we should stay on the safe side and at least check the version and possibly vendor.

Think about some strange corner cases, e.g. the SLES for SAP installation medium includes the original SLES product and additionally the SLES-for-SAP product, i.e. there are actually two (!) products there...

test/pkg_finish_test.rb
```diff
@@ -0,0 +1,182 @@
...
+        it "does not disabled those repositories" do
+          expect(cd_repo).to_not receive(:disable!)
+          client.run
+        end
+      end
+    end
+
+    context "during update" do
+      let(:update) { true }
+      let(:tmpdir) { Pathname.new(__FILE__).dirname.join("tmp") }
```
lslezak: NP: maybe use a real tmp dir created by `Dir.mktmpdir`?....

test/pkg_finish_test.rb
```diff
@@ -0,0 +1,182 @@
...
+
+      before do
+        allow(Yast::Directory).to receive(:vardir).and_return(vardir.to_s)
+        allow(Yast::WFM).to receive(:call)
+        stub_const("Yast::PkgFinishClient::REPOS_DIR", repos_dir.to_s)
+        stub_const("Yast::Pkg", double("pkg").as_null_object)
+      end
+
+      around do |example|
+        FileUtils.rm_rf(tmpdir)
```
lslezak: NP: ... then you could remove this cleanup at the beginning, the `mktmpdir` ensures it's a new dir (without any possible leftovers from previous tests or runs).

src/lib/packager/clients/pkg_finish.rb
imobach: Yes, that was the reason to have `ProductRepository#==` implemented. I've fixed it to rely on it.

src/lib/packager/clients/pkg_finish.rb
kobliha: If there is some docu or code that CAN be linked, I'd love to see the reference here as a comment in the code. Please :)

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,171 @@
...
+
+      cmd = WFM.Execute(path(".local.bash_output"), "date +%Y%m%d-%H%M%S")
+      a_name_list = (cmd["stdout"] || "the_latest").split("\n")
+      archive_name = "repos_#{a_name_list.first}.tgz"
+
+      compress_cmd = format(TAR_CMD,
+        source: String.Quote(Ops.add(Directory.vardir, "/repos.d_backup/")),
+        archive: String.Quote(archive_name),
+        target: String.Quote(REPOS_DIR))
+#
```
kobliha: What's this? Missing a comment?

kobliha:
It finally looks like a Ruby code after your refactoring :) Nice!

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,198 @@
...
+module Yast
+  class PkgFinishClient < ::Installation::FinishClient
+    include Yast::I18n
+    include Yast::Logger
+
+    # Path to libzypp repositories
+    REPOS_DIR = "/etc/zypp/repos.d"
+    # Repository schemes considered local (see disable_local_repos)
+    # https://github.com/openSUSE/libzypp/blob/a7a038aeda1ad6d9e441e7d3755612aa83320dce/zypp/Url.cc#L458
+    LOCAL_SCHEMES = [:cd, :dvd, :dir, :hd, :iso, :file]
```
imobach: That's the wrong approach. I'll move this logic to a `Packages::Repository#local?` method.

lslezak:
Note: The Travis failure is caused by the dependency on https://github.com/yast/yast-yast2/pull/465, it needs to be merged first.

package/yast2-packager.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue May 10 12:31:13 UTC 2016 - igonzalezsosa@suse.com
+
+- Local repositories are disabled at the end of the installation
+  if they're not needed anymore (FATE#320494)
```
lslezak: The `not need need anymore` part is a bit vague and unclear, I'd write something like *if there is an online repository replacement available*  or something like that.

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -0,0 +1,192 @@
...
+    # @return [String] Name of the backup archive (locate in the given target directory)
+    def backup_old_sources(source, target)
+      archive_name = "repos_#{Time.now.strftime(BACKUP_TIMESTAMP_FORMAT)}.tgz"
+      compress_cmd = format(TAR_CMD,
+        target: String.Quote(target),
+        archive: archive_name,
+        source: String.Quote(source))
+      cmd = SCR.Execute(path(".target.bash_output"), compress_cmd)
+      if !cmd["exit"].zero?
+        log.error("Unable to backup current repos; Command >#{compress_cmd}< returned: #{cmd}")
```
lslezak: If it fails then the returned value is the result of `log.error` which is wrong. It should either throw an exception or return `nil` on error.

test/pkg_finish_test.rb
```diff
@@ -0,0 +1,187 @@
...
+require "fileutils"
+
+describe Yast::PkgFinishClient do
+  Yast.import "Pkg"
+  Yast.import "Installation"
+  Yast.import "WFM"
+
+  FAILED_PKGS_PATH = "/var/lib/YaST2/failed_packages"
+
+  subject(:client) { Yast::PkgFinishClient.new }
```
lslezak: IIRC the default `subject` is an instance of the class described in the `describe` block. I think you can remove this line and use `subject` instead of `client` everywhere.

test/pkg_finish_test.rb
```diff
@@ -0,0 +1,187 @@
...
+      expect(Yast::Pkg).to receive(:SourceSaveAll)
+      expect(Yast::Pkg).to receive(:TargetFinish)
+      expect(Yast::Pkg).to receive(:SourceCacheCopyTo).with(destdir)
+      allow(Yast::WFM).to receive(:Execute)
+      expect(client.run).to be_nil
+    end
+
+    it "copies failed_packages list under destination dir" do
+      stub_const("Yast::Pkg", double("pkg").as_null_object)
+      allow(::File).to receive(:exist?).and_call_original
```
lslezak: You do not need to qualify the `File` class here as the code is not nested in any `module`.

test/pkg_finish_test.rb
```diff
@@ -0,0 +1,187 @@
...
+      expect(::File).to receive(:exist?).with(FAILED_PKGS_PATH)
+        .and_return(true)
+      expect(::FileUtils).to receive(:cp)
```
lslezak: The same here, moreover a plain `FileUtils` is used below so it is even inconsistent...

test/pkg_finish_test.rb
imobach: You're right. But to be honest I usually prefer the use of [named subjects](http://blog.davidchelimsky.net/blog/2012/05/13/spec-smell-explicit-use-of-subject/).

lslezak:
LGTM

imobach:
Thanks. Travis ss finally green so I'll merge now.

PR #162, opened 2016-06-09
Obsolete devel-doc package

imobach:


package/yast2-packager.spec
```diff
@@ -83,6 +83,8 @@ Recommends:     yast2-registration
 # force *-webpin subpackage removal at upgrade
 Obsoletes:      yast2-packager-webpin < %version
```
mchf: NP: why empty line?

imobach: To make clear that it has nothing to do with the `-webpin` thingie (although is obvious).

mchf:
LGTM

imobach:
Thanks!

PR #164, opened 2016-06-15
Add ssl_verify=no to InstURL.installInf2Url when needed

imobach:
Fixes [bsc#982727](https://bugzilla.suse.com/show_bug.cgi?id=982727)


src/modules/InstURL.rb
```diff
@@ -142,232 +83,45 @@ def RewriteCDUrl(url)
...
-          )
-          repo_url_from_inf = true
-          raise Break
-        end
-      end
+    def installInf2Url(extra_dir = "")
+      return @installInf2Url unless @installInf2Url.nil?
 
-      return repo_url if repo_url_from_inf
+      parts = URL.Parse(Linuxrc.InstallInf("ZyppRepoURL"))
```
lslezak: NTH: I'd prefer native Ruby `URI` here, in the log term we should get rid of `Yast::URL`...

src/modules/InstURL.rb
```diff
@@ -142,232 +83,45 @@ def RewriteCDUrl(url)
...
-      Builtins.y2debug("URL %1", URL.HidePassword(url))
-      url
+    # Determines whether the installation URL is remote or not
+    #
+    # @return [Boolean] true if it's remote; false otherwise.
+    # @see installInf2Url
+    def is_network
+      return @is_network unless @is_network.nil?
+      scheme = URL.Parse(installInf2Url("")).fetch("scheme")
+      @is_network = !LOCAL_SCHEMES.include?(scheme)
```
lslezak: NTH: I'd use `scheme.downcase` here, just to be more error proof when somebody writes `INSTALL=HTTP://EXAMPLE.COM`...

package/yast2-packager.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jun 15 11:07:50 UTC 2016 - igonzalezsosa@suse.com
+
+- Allow installation via HTTPS using a self-signed certificate
```
lslezak: I'd add `... when "ssl_verify=no" boot option is used`.

Just to be clear that we by default do not ignore certificate issues, someone might get scared about a possible security issue... :wink:

src/modules/InstURL.rb
imobach: I wrote it using the native Ruby `URI` but I changed to `Yast::URL` because of this:

```ruby
url = "cd:///?device=/dev/sda"
URI(url).to_s #=> "cd:/?device=/dev/sda"
Yast::URL.Build(Yast::URL.Parse(url)) #=> "cd:///?device=/dev/sda"
```

If that's not a problem, I'm fine with changing it to URI.

lslezak:
BTW thanks for the clean up and dropping the old tests! :+1:

src/modules/InstURL.rb
lslezak: Oh, I see. Then I'm fine with `Yast::URL`, just mention this issue in a comment so we know this issue later when somebody would like to do the switch.

imobach: Ok, no problem, I'll add a comment then.

imobach: Oh, I've checked Linuxrc docs and ZyppRepoURL from a SP2 inst-sys and they use `cd:/?device=...`. So I guess we're in the safe side and I was confused by old examples.

mvidner: I wanted to check the documentation of install.inf contents, but I could not find it at http://www.rubydoc.info/github/yast/yast-yast2/Yast/LinuxrcClass#InstallInf-instance_method . It should link to the following, I guess:

I did find https://en.opensuse.org/SDB:Linuxrc_install.inf but it does not document `ZyppRepoURL`, only

> xxxZyppRepoURL 	String 	Not in use yet. After renaming to ZyppRepoURL will contain libzypp installation repository URL. 

@wfeldt Steffen, is there a place documenting this which I overlooked?

wfeldt: > Steffen, is there a place documenting this which I overlooked?

Definitely.

https://github.com/openSUSE/linuxrc/blob/master/linuxrc_yast_interface.txt

test/inst_url_test.rb
```diff
@@ -0,0 +1,102 @@
...
+  subject(:inst_url)  { Yast::InstURL }
+    
+  describe "#installInf2Url" do
+    let(:zypp_repo_url) { "http://opensuse.org/repo" }
+    let(:ssl_verify) { "yes" }
+
+    before do
+      Yast::InstURL.main
+      allow(Yast::Linuxrc).to receive(:InstallInf).with("ZyppRepoURL")
+        .and_return(zypp_repo_url ? zypp_repo_url.clone : nil)
```
mvidner: Why do we need to clone the string? I think it is OK to mutate strings inside methods as an implementation detail, but for API purposes we should treat them as immutable. (Ruby 3 may make them so).

src/modules/InstURL.rb
```diff
@@ -142,232 +83,45 @@ def RewriteCDUrl(url)
...
-          end
-
-          slash = ""
-          slash = "/" if Builtins.substring(extra_dir, 0, 1) != "/"
-          serverdir = Ops.add(Ops.add(serverdir, slash), extra_dir)
-          slash = ""
-        else
-          serverdir = extra_dir
-        end
+      if parts["scheme"] == "https" && !SSLVerificationEnabled()
```
mvidner: `SSLVerificationEnabled` checks whether `Linuxrc.InstallInf("ssl_verify") != "no"` but ssl_verify is not documented in https://en.opensuse.org/SDB:Linuxrc_install.inf so there are potential bugs when checking/using "0" or "false".

test/inst_url_test.rb
imobach: Yes, that comes from my first attempt while writing those tests. It's not needed anymore. Thanks for pointing it out.

src/modules/InstURL.rb
wfeldt: @mvidner That's not possible. Please check the docs.

src/modules/InstURL.rb
```diff
@@ -142,232 +83,45 @@ def RewriteCDUrl(url)
...
+      parts = URL.Parse(Linuxrc.InstallInf("ZyppRepoURL"))
 
-      # Initial repository hasn't been found in install.inf
-      # Trying to guess it ...
-      Builtins.y2warning("Initial repository not found in install.inf")
-
-      options = ""
-      url_tokens = {}
-
-      instmode = Linuxrc.InstallInf("InstMode") # mode
```
mvidner: `InstMode` is still documented in https://en.opensuse.org/SDB:Linuxrc_install.inf

imobach: It looks like the [SDB page](https://en.opensuse.org/SDB:Linuxrc%20install.inf) is quite outdated. I guess we should update that page using https://github.com/openSUSE/linuxrc/blob/master/linuxrc_yast_interface.txt as reference. I can take care if @mvidner and @wfeldt agree.

src/modules/InstURL.rb
```diff
@@ -86,24 +87,27 @@ def RewriteCDUrl(url)
...
+      if zypp_repo_url.empty?
         # Make it compatible with the current behaviour when
         # install.inf does not exist.
         log.warn "No URL specified through ZyppRepoURL"
-        parts = URL.Parse("cd:///")
+        zypp_repo_url = "cd:/"
       end
 
-      parts["path"] = File.join(parts["path"], extra_dir) unless extra_dir.empty?
+      uri = URI(zypp_repo_url)
```
mvidner: I disagree with this. URI parsing is a minefield, and this change will interpret exotic URIs in a slightly different way than before, or throw exceptions.

src/modules/InstURL.rb
wfeldt: Sure, go ahead. I wasn't even aware of this SDB page until today..

src/modules/InstURL.rb
imobach: Well, maybe we could postpone this change when SP2 is already released.

lslezak: I think Martin's point is that we parse and then build the URL back. That might not produce the very same string in some edge cases. We should return the original string and do the modification *only* when really need (HTTPS *and* `ssl_verify=no` used).

mvidner:
I have added an obsolescence warning to https://en.opensuse.org/SDB:Linuxrc_install.inf#Linuxrc_Info_File:_install.inf , pointing to https://github.com/openSUSE/linuxrc/blob/master/linuxrc_yast_interface.txt

src/modules/InstURL.rb
lslezak: And even the modification should be probably done without parsing, something like
```ruby
url += (parts["query"].empty? ? "?" : "&") + "ssl_verify=no"
```
to keep the original string as much as possible.

imobach: From my POV, that would make code harder to read and extend. Now we have two different cases to modify the URL:

* Adding an `extra_dir`.
* Skipping SSL verification.

If we add another case in the future, you'll need to parse the URL a third time. From my POV, the current implementation is more straightforward. Anyway, I see Martin's point.

So I'm fine with changing it to something like (names may change):

```ruby
zypp_repo_url = add_extra_dir(zypp_repo_url, extra_dir)
zypp_repo_url = add_ssl_verify(zypp_repo_url)
```

lslezak: The proposal LGTM.

I know that working with the parsed URL is simpler and can be easily extended, but libzypp sometimes uses quite complicated URLs (`iso://...`, `smb://...`) and the risk of building the URL back wrongly is quite high. In this case I'd rather stay on the safe side and sacrifice the better code readability.

mvidner: It is a tradeoff between old compatibility custom libraries and the standard Ruby libraries. I don't want to hold us back forever. We are using `URI` in some places anyway: https://github.com/search?utf8=%E2%9C%93&q=uri+org%3Ayast&type=Code&ref=searchresults
So to clarify my original point: I disagree with the change being introduced quietly. If we mention it in the changelog ("use URI.parse on ZyppRepoURL") and we are aware of the (small?) risk then it is OK.

More reference: https://en.opensuse.org/openSUSE:Libzypp_URIs

imobach: @lslezak @mvidner Well, the other option is just use the `Yast::URL` module (as the original PR) and leave the switch for the future.

lslezak: Well, I'd still prefer not parsing the URL unless really needed. The original code simply returned the `Linuxrc.InstallInf("ZyppRepoURL")` value.

The building URL code following later is actually not used for some time, it's dead code. It was kept just for compatibility reasons (IIRC some openSUSE release switched to `ZyppRepoURL` but the paralel SLE was still building the parts. Now we use `ZyppRepoURL` everywhere, we can drop that URL build part.)

src/modules/InstURL.rb
```diff
@@ -142,232 +83,68 @@ def RewriteCDUrl(url)
...
-            url = Ops.add(Ops.add(url, option_separator), "ssl_verify=no")
-            option_separator = "&"
-          end
-        end
-      end # is_network
+    # Determines whether the installation URL is remote or not
+    #
+    # @return [Boolean] true if it's remote; false otherwise.
+    # @see installInf2Url
+    def is_network
```
lslezak: I haven't found any usage of `InstURL.is_network` in whole YaST and in this module it is not used as well. Maybe we could drop it??

src/modules/InstURL.rb
```diff
@@ -142,232 +83,68 @@ def RewriteCDUrl(url)
...
+      return @is_network unless @is_network.nil?
+      scheme = URL.Parse(installInf2Url("")).fetch("scheme")
+      @is_network = !LOCAL_SCHEMES.include?(scheme.downcase)
```
lslezak: NP: The things get more tricky with the `iso` URL scheme, e.g. `iso://?iso=image.iso&url=file://...` is local while `iso://?iso=image.iso&url=nfs://...` is not. But considering that the method is not used anymore let's ignore it...

src/modules/InstURL.rb
```diff
@@ -142,232 +83,68 @@ def RewriteCDUrl(url)
...
+    # Helper method to add ssl_verify parameter if needed
+    #
+    # Only applicable if scheme is 'https'.
+    #
+    # @param url       [String] URL
+    # @return [String] URL (with ssl_verify set to 'no' if needed)
+    def add_ssl_verify_no_to_url(url)
+      parts = URL.Parse(url)
+      return url if parts["scheme"].downcase != "https"
+      log.error "Disabling certificate check for the installation repository"
```
lslezak: It's not an error, the disabling is done on the user request, `log.warn` is more appropriate IMHO.

src/modules/InstURL.rb
imobach: I'm fine with dropping it. I just wanted to keep the API but I'm pretty sure that it's not used anymore. And about 3th party YaST modules I would be suprissed if it's used.

src/modules/InstURL.rb
imobach: I agree. But I found [this](https://github.com/yast/rubocop-yast/blob/0cdddc26bead94730322dbf4d99f8002318082f0/lib/rubocop/yast/builtins/y2log.rb#L15).

lslezak:
My last objection: Please mention the `InstURL.is_network` removal in the `*.changes` file.

lslezak:
LGTM :tada: 

Thanks! :+1:

imobach:
Thanks for the review.

PR #167, opened 2016-06-23
The boot option to allow self-signed certs is ssl_certs=0.

mvidner:
If ssl_certs=0 appears on the command line, ssl_verify=no appears in install.inf.

With this, I have tested that the #164 fix actually works, and in the process found a one-liner for an HTTPS server: 
```
ruby -r webrick/https -e 'WEBrick::HTTPServer.new(Port:8000,DocumentRoot:".",SSLEnable:true,SSLCertName:[%w[CN localhost]]).start'
```


lslezak:
LGTM, thanks for testing and fixing! :+1:

PR #171, opened 2016-07-13
 KDE don't know how to open an RPM file (boo#954143)

marceloatie:
KDE don't know how to open an RPM file (boo#954143)


jreidinger:
do you test modified code? I do not use KDE, so I cannot test locally

jreidinger:
thanks for patch, will merge it and add changes mentioning you

PR #173, opened 2016-07-31
Prioritize yast-packager for opening rpms

marceloatie:
Added _InitialPreference=10_ in yast2-packager.desktop

### DIFF
--- a/desktop/yast2-packager.desktop
+++ b/desktop/yast2-packager.desktop
@@ -8,6 +8,7 @@ Terminal=false
 Type=Application
 Categories=System;PackageManager;X-SuSE-ControlCenter-System;
 MimeType=application/x-rpm;application/x-redhat-package-manager;
+InitialPreference=10
 NotShowIn=GNOME;MATE;
 StartupNotify=true
 
--- a/desktop/yast2-packager.desktop
+++ b/desktop/yast2-packager.desktop
@@ -8,7 +8,9 @@ Terminal=false
 Type=Application
 Categories=System;PackageManager;X-SuSE-ControlCenter-System;
 MimeType=application/x-rpm;application/x-redhat-package-manager;
-InitialPreference=10
 NotShowIn=GNOME;MATE;
 StartupNotify=true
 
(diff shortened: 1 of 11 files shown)


package/yast2-packager.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Sun Jul 31 13:47:36 UTC 2016 - marceloatie@gmail.com
+
+- Prioritize yast-packager for opening rpms
```
lslezak: Any bugzilla bug number for reference? Without it the OBS will complain...

desktop/yast2-packager.desktop
```diff
@@ -8,6 +8,7 @@ Terminal=false
 Type=Application
 Categories=System;PackageManager;X-SuSE-ControlCenter-System;
 MimeType=application/x-rpm;application/x-redhat-package-manager;
+InitialPreference=10
```
lslezak: The [freedesktop documentation](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#kde-items) says that this keyword is KDE specific. Please add a comment with the documentation link and description that this is only for KDE and does not affect the other desktop environments.

lslezak:
BTW there is a conflict in Git, please resolve it (cannot be merged).

Otherwise the change looks good, thank you for your contribution!

marceloatie:
Bugzilla number is boo#954143

lslezak:
I can see a fix for SLE12-SP1 (https://github.com/yast/yast-packager/pull/174), do we want to have it fixed in SP2 as well? If yes, the correct workflow is to fix it in the SP2 branch first and then merge the SP2 branch into `master`.

With this workflow we can easily check whether all SP2 fixes are present also in `master` (so they are also included in the future SP3). See more details [here](http://yastgithubio.readthedocs.io/en/latest/maintenance-branches/#maintenance-work-flow).

lslezak:
This fix is included in #195, let's close this duplicate...

PR #174, opened 2016-07-31
Allow KDE to use packager for opening rpms

marceloatie:
Allow KDE to use packager for opening rpms and prioritize over other options.

### DIFF
--- a/desktop/yast2-packager.desktop
+++ b/desktop/yast2-packager.desktop
@@ -8,6 +8,7 @@ Terminal=false
 Type=Application
 Categories=System;PackageManager;X-SuSE-ControlCenter-System;
 MimeType=application/x-rpm;application/x-redhat-package-manager;
-NotShowIn=KDE;GNOME;MATE;
+InitialPreference=10
+NotShowIn=GNOME;MATE;
 StartupNotify=true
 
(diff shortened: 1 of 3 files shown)


desktop/yast2-packager.desktop
```diff
@@ -8,6 +8,7 @@ Terminal=false
 Type=Application
 Categories=System;PackageManager;X-SuSE-ControlCenter-System;
 MimeType=application/x-rpm;application/x-redhat-package-manager;
-NotShowIn=KDE;GNOME;MATE;
+InitialPreference=10
+NotShowIn=GNOME;MATE;
```
lslezak: Sorry, this line cannot go to SP1 as SLES supports only Gnome. This change would cause a regression there..

lslezak:
And the same comments as in https://github.com/yast/yast-packager/pull/173 applies here as well...

desktop/yast2-packager.desktop
marceloatie: Gnome will not be affected, just KDE. (I tested on openSUSE)

lslezak:
Closing as SP1 is too old now, this is being solved in #190 and #195 for SP2/SP3.

PR #175, opened 2016-08-09
More robust unit tests

ancorgs:
Found by accident. The previous version breaks if the system running the test has a /var/lib/YaST2/failed_packages file.


lslezak:
LGTM

PR #176, opened 2016-08-09
Remove old libstorage dependency

ancorgs:
To make travis work again we need to be able to build and install some basic packages without bringing yast2-storage in. We don't need all the code to actually work, just the one executed in our installation.

This pull request removes the dependency from the spec file and temporarily comments some test to make the build process work. Commented code is being documented at https://github.com/yast/yast-storage-ng/blob/master/doc/installer-hacks.md


ancorgs:
This should not target master. Closing.

PR #177, opened 2016-08-09
Remove old libstorage dependency

ancorgs:
To make travis work again we need to be able to build and install some basic packages without bringing yast2-storage in. We don't need all the code to actually work, just the one executed in our installation.

This pull request removes the dependency from the spec file and temporarily comments some test to make the build process work. Commented code is being documented at https://github.com/yast/yast-storage-ng/blob/master/doc/installer-hacks.md

### DIFF
--- a/test/space_calculation_test.rb
+++ b/test/space_calculation_test.rb
@@ -112,6 +112,8 @@ describe Yast::SpaceCalculation do
     end
   end
 
+# storage-ng
+=begin
   describe "#size_from_string" do
     it "converts string without units bytes" do
       expect(Yast::SpaceCalculation.size_from_string("42.00")).to eq(42)
@@ -153,6 +155,7 @@ describe Yast::SpaceCalculation do
       expect(Yast::SpaceCalculation.size_from_string("0.00")).to eq(0)
     end
   end
+=end
 
   describe "#btrfs_snapshots?" do
     let(:dir) { "/mnt" }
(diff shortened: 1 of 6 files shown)


test/space_calculation_test.rb
```diff
@@ -112,6 +112,8 @@ def expect_to_execute(command)
     end
   end
 
+# storage-ng
+=begin
```
lslezak: NP: Um, personally I do not like these block comments, it breaks simple grepping the sources when you are looking for something, it might confuse some simple syntax highlighters and you can easily overlook that the block is commented when the block it too long.

I'd prefer the usual '#' comments, editors/IDEs usually have some shortcuts for (un)commenting multiple lines.

ancorgs: Yes, they are unusual and that's exactly why are we using it to comment temporary commented code. This format makes very easy to find the temporary patches.
```ruby
  # storage-ng
  value = "replaced value because of storage-ng"
=begin
  value = "something"
  value << "original that we need to comment"
  value << "and that ends here"
=end
```

These comments are not expected to be there after the whole rewrite. That's what I want them to look different.

ancorgs: I will add a comment to https://github.com/yast/yast-storage-ng/blob/master/doc/installer-hacks.md explaining the rational in the next update (due today).

lslezak:
Ok, given the comments are temporary LGTM.

lslezak:
Ok, LGTM

PR #178, opened 2016-08-10
[Review] Request from 'schubi2' @ 'yast/yast-packager/review_160810_add_on_product_license'

schubi2:
I have tested the latest build together the latest SDK release.

  * f1c7e9c Do not ask add-on product licences again after restarting YaST due self-update

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 10 10:33:50 CEST 2016 - schubi@suse.de
+
+- Do not ask add-on product licences again after restarting
+  YaST due self-update. (bnc#992608)
+- 3.1.110
+
+-------------------------------------------------------------------
 Thu Jul 28 10:40:32 UTC 2016 - jreidinger@suse.com
 
 - allow KDE to use packager for opening rpms (boo#954143)
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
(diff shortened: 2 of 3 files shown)


kobliha:
@schubi2 And what happens if the user actually `declined` the license? IMO it was not added, but then you will add it without asking...

imobachgs:
Moreover we need to check if this method is used also when adding regular add-ons (not added using a `add_on_products.xml`) (I'm not sure about it).

schubi2:
@kobliha 
"And what happens if the user actually declined the license? IMO it was not added, but then you will add it without asking..." I have tried: The user cannot continue without accepting this add-on license.

schubi2:
I have tested some use cases documented here:
https://trello.com/c/h7DzsthA/647-2-sle-12-sp2-p1-992608-add-ons-added-via-add-on-products-xml-are-lost-after-self-update
@kobliha , @imobachgs have I missed some other use cases ?

imobachgs:
I've found a scenario that won't work (note `<ask_user>` option):

```xml
<add_on_products>
  <product_items config:type="list">
    <product_item>
      <name>SUSE Linux Enterprise Software Development Kit 12</name>
      <url>cd:/?alias=SDK</url>
      <path>/</path>
      <priority config:type="integer">50</priority>
      <ask_user config:type="boolean">true</check_media>
    </product_item>
  </product_items></add_on_products>
```

The user will be asked twice whether to include the add-on or not.

lslezak:
JFYI: we should not need this after moving the self update step earlier.

schubi2:
It is obsolete.

PR #179, opened 2016-08-10
Fix add_on_products.xml handling

imobachgs:
Fixes [bsc#991935](https://bugzilla.suse.com/show_bug.cgi?id=991935):

* Fix endless loop.
* Add a `<check_name>` option to disable the `<name>` tag check.
* User won't be asked to insert a CD/DVD is a valid repo is found.


wfeldt:
This fixes part of the issues raised in the mentioned bug (url parsing, the infinite dialog loop). As long as it's understood that more fixes are needed, lgtm.

wfeldt:
lgtm

PR #180, opened 2016-08-11
Sorted addons below product base (part of bsc#992304)

teclator:
### Changes
-  Addons are sorted based on "source" key, that basically represent the order in which the repositories where added to the system, so the base product will be the one with the lower "source" and for that the first one listed.
- 3.1.110

### Test
- Added unit tests and tested manually in a SLE 12 - SP2 - Server,  build 2050.

#### Original State

Addons are listed unordered and product base is not the first one listed.

beforefix

Addons are listed ordered being the product base the first one of the list.

fixed

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -16,6 +16,10 @@ module Yast
 
     # All known types of resolvables
     RESOLVABLE_TYPES = [:product, :patch, :package, :pattern, :language]
+
+    # Key to sort by resolvable selection
+    RESOLVABLE_SORT_ORDER = { :product => "source", :pattern => "order" }
+
     # Minimum set of packages tags required to enable VNC server
     VNC_BASE_TAGS = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts", "xinetd"]
     # Additional packages tags needed to run second stage in graphical mode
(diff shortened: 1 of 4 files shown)


test/packages_test.rb
```diff
@@ -897,4 +897,92 @@ def product(properties = {})
...
+    it "filters not user visible resolvables from the list for type pattern" do
+      expect(Yast::Pkg).to receive(:ResolvableProperties).with("", :pattern, "")
+        .and_return(unordered_packages)
+      expect(subject).to receive(:sort_resolvable!)
+        .with(filtered_packages, :pattern)
+
+      subject.ListSelected(:pattern, "")
+    end
+
+    it "sorts resultant list depending on resortable type" do
```
lslezak: Typo: *resolvable type*

teclator: Ups :)

src/modules/Packages.rb
```diff
@@ -2599,6 +2578,35 @@ def log_resolvables(text, resolvables)
       end
     end
 
+    # Prepares a list of formatted selected resolvables
+    #
+    # :pattern resolvables are sorted by "order"
```
lslezak: These two lines should be removed, this function does not sort.

test/packages_test.rb
```diff
@@ -897,4 +897,92 @@ def product(properties = {})
...
+    end
+
+    let(:ordered_products) do
+      [
+        product("name" => "p1", "status" => :selected, "source" => 10),
+        product("name" => "p3", "status" => :selected, "source" => 15)
+      ]
+    end
+
+    let(:unordered_packages) do
```
lslezak: This should be obviously named `:unordered_patterns`.

test/packages_test.rb
```diff
@@ -897,4 +897,92 @@ def product(properties = {})
...
+      [
+        pattern("name" => "p3", "status" => :selected, "order" => "3", "user_visible" => true),
+        pattern("name" => "p1", "status" => :selected, "order" => "1", "user_visible" => false),
+        pattern("name" => "p2", "status" => :available, "order" => "2", "user_visible" => true)
+      ]
+    end
+
+    let(:filtered_packages) do
```
lslezak: And `:filtered_patterns` here.

lslezak:
BTW thanks for the nice test suite! :thumbsup:

teclator:
Thnx to you for the review i also tried to use your same format to summarize the PR, updating it..

PR #181, opened 2016-08-15
parameter error

schubi2:


src/modules/AddOnProduct.rb
```diff
@@ -1579,7 +1579,7 @@ def AddRepo(url, pth, priority)
       log.info "New repository id: #{new_repo_id}"
 
       if new_repo_id == nil || new_repo_id < 0
-        log.error "Unable to add product: %1", URL.HidePassword(url)
+        log.error "Unable to add product: #{URL.HidePassword(url)}"
```
mvidner: NP: Actually couldn't this be deleted completely since `Report.Error` logs the reported message too?
https://github.com/yast/yast-yast2/blob/e5dc920a7b436994ce5f3ba20be21917136b2118/library/general/src/modules/Report.rb#L526

PR #182, opened 2016-08-16
Fixed required acceptance of license in installation

kobliha:
- When tarball with licenses contains no-acceptance-needed file, explicit acceptance of that license is not required. This did not work after refactoring of inst_complex_welcome as the installer called AcceptanceNeeded, but data were not loaded yet (bnc#993285)
- The fix adds lazy loading into AcceptanceNeeded function, but keeps it also on the old place
- This only applies for the base product license in the initial installation

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -311,11 +311,13 @@ module Yast
     # If no sources are found, returns 0.
     # FIXME: Connected to bsc#993285, refactoring needed
     #
-    # return [Integer] base_product_id or nil
+    # return [Integer] base_product_id or 0
     def base_product_id
       raise "Base product can be only found in installation" unless Stage.initial
 
       # The first product in the list of known products
+      # 0 is the backward-compatible default value, first installation repo always
+      # gets this ID later
       current_sources = Pkg.SourceGetCurrent(true)
       current_sources.any? ? current_sources.first : 0
     end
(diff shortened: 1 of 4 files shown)


src/modules/ProductLicense.rb
```diff
@@ -302,12 +307,48 @@ def GetLicenseDialogTerm(languages, license_language, licenses, id)
...
+    # If no sources are found, returns 0.
+    # FIXME: Connected to bsc#993285, refactoring needed
+    #
+    # return [Integer] base_product_id or nil
+    def base_product_id
+      raise "Base product can be only found in installation" unless Stage.initial
+
+      # The first product in the list of known products
+      current_sources = Pkg.SourceGetCurrent(true)
+      current_sources.any? ? current_sources.first : 0
```
mchf: according doc it should return base_product or nil. So, I guess that ```s/nil/zero/``` in docu.

I'm also curios if ```0``` is somehow special product id ... or why it is used as default.

src/modules/ProductLicense.rb
```diff
@@ -302,12 +307,48 @@ def GetLicenseDialogTerm(languages, license_language, licenses, id)
...
     # @see BNC #448598
     # @return [Boolean] if required
-    def AcceptanceNeeded(id)
-      Ops.get(@license_acceptance_needed, id, true)
+    def AcceptanceNeeded(src_id)
+      # FIXME: lazy loading of the info about licenses, bigger refactoring needed
+      # @see bsc#993285
+      #
+      # In the initial installation, for base product, acceptance_needed needs
+      # to be know before showing the license dialog (inst_complex_welcome).
```
mchf: ```known```

src/modules/ProductLicense.rb
```diff
@@ -302,12 +307,48 @@ def GetLicenseDialogTerm(languages, license_language, licenses, id)
...
+    # Returns source ID of the base product - initial installation only!
+    # If no sources are found, returns 0.
+    # FIXME: Connected to bsc#993285, refactoring needed
+    #
+    # return [Integer] base_product_id or nil
+    def base_product_id
+      raise "Base product can be only found in installation" unless Stage.initial
+
+      # The first product in the list of known products
+      current_sources = Pkg.SourceGetCurrent(true)
```
mchf: I assume that ```Pkg.SourceGetCurrent``` cannot return ```nil```.

src/modules/ProductLicense.rb
```diff
@@ -793,20 +835,21 @@ def GetSourceLicenseDirectory(src_id, fallback_dir)
...
-      if FileUtils.Exists(
-          Builtins.sformat("%1/no-acceptance-needed", @license_dir)
-        )
-        if id == nil
-          Builtins.y2error("Parameter id not set")
-        else
-          SetAcceptanceNeeded(id, false)
-        end
-      end
+      cache_license_acceptance_needed(src_id, @license_dir)
```
mchf: both ```src_id``` and ```id``` exists as param. Dropped code uses ```id``, new code uses ```src_id``` So which one is the proper one?

src/modules/ProductLicense.rb
kobliha: Ah, `0` is the correct backward-bug-compatible default value

test/product_license_test.rb
```diff
@@ -147,4 +147,101 @@
...
+      # and thus we need to reinit all caches for testing with different values
+      Yast::ProductLicense.initialize_default_values
+
+      allow(Yast::ProductLicense).to receive(:base_product_id).and_return(base_product_id)
+    end
+
+    context "when called in the initial stage of installation" do
+      before do
+        # Initial installation
+        allow(Yast::Stage).to receive(:stage).and_return("initial")
```
mchf: mocking internals of other module.

In my POV proper mock should be ```allow(Stage).to receive(:initial).and_return(true)```

test/product_license_test.rb
```diff
@@ -147,4 +147,101 @@
...
+        allow(Yast::Mode).to receive(:mode).and_return("installation")
```
mchf: Similar as above.

src/modules/ProductLicense.rb
kobliha: Empty list, if nothing is found

test/product_license_test.rb
```diff
@@ -147,4 +147,101 @@
...
+            expect(Yast::ProductLicense.AcceptanceNeeded(add_on_product_id)).to eq(false)
+          end
+        end
+      end
+    end
+
+    context "when called on a running system" do
+      before do
+        allow(Yast::Stage).to receive(:stage).and_return("normal")
+        allow(Yast::Mode).to receive(:mode).and_return("normal")
```
mchf: similar as above. I guess you need check if running in installer or not, so mocking ```false``` for ```Stage.initial``` should make it.

src/modules/ProductLicense.rb
kobliha: OK :)

test/product_license_test.rb
kobliha: c'est possible

test/product_license_test.rb
kobliha: OK

src/modules/ProductLicense.rb
```diff
@@ -320,15 +367,12 @@ def SetAcceptanceNeeded(id, new_value)
         return
       end
 
-      Ops.set(@license_acceptance_needed, id, new_value)
+      @license_acceptance_needed[id] = new_value
 
       if new_value == true
```
mchf: NP: ```if new_value``` is enough ;-)

mchf:
lgtm

kobliha:
@mchf Thanks for review

PR #183, opened 2016-08-16
Remove the icon in the add-on selection popup (bsc#875201)

lslezak:
- All icons in all popups should be removed, see https://bugzilla.suse.com/show_bug.cgi?id=875201#c1
- 3.1.114

## Test

Tested manually with this simple `add_on_products.xml` file:

```xml
<?xml version="1.0"?>
<add_on_products xmlns="http://www.suse.com/1.0/yast2ns"
    xmlns:config="http://www.suse.com/1.0/configns">
    <product_items config:type="list">
        <product_item>
            <name>Driver Update Add On</name>
            <url>cd:///?alias=dud</url>
            <path>/dud</path>
            <priority config:type="integer">50</priority>
            <ask_user config:type="boolean">true</ask_user>
        </product_item>
    </product_items>
</add_on_products>
```

using this simple Ruby script:

```ruby
require "yast"
Yast.import "AddOnProduct"
Yast::AddOnProduct.ParseXMLBasedAddOnProductsFile("add_on_products.xml", "cd:/")
```

### Screenshots

*(Taken in an installed system, the colors will be of course different in a real SLE12 installation...)*

#### The Original State

yast2-addon_selection_orig

#### After Icon Removal

yast2-addon_selection_fixed

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Aug 16 14:16:01 UTC 2016 - lslezak@suse.cz
+
+- Remove the icon from the add-on selection popup (bsc#875201)
+- 3.1.114
+
+-------------------------------------------------------------------
 Tue Aug 16 08:31:23 CEST 2016 - locilka@suse.com
 
 - Added lazy loading of licenses into AcceptanceNeeded in case of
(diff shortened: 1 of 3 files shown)


mvidner:
LGTM

kobliha:
@lslezak @mvidner Thanks for taking care! ;)

PR #184, opened 2016-08-18
do not raise if no license has to be accepted

schubi2:
https://bugzilla.suse.com/show_bug.cgi?id=994309

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -2,7 +2,7 @@
 Thu Aug 18 10:55:06 CEST 2016 - schubi@suse.de
 
 - Do not raise an exception if no license is available (no licence
-  dir at all) to accept. This is still a fix for bnc#993285.
+  dir at all) to accept (bnc#994309).
 - 3.1.115
 
 -------------------------------------------------------------------
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 18 10:55:06 CEST 2016 - schubi@suse.de
+
+- Do not raise an exception if no license is available (no licence
+  dir at all) to accept. This is still a fix for bnc#993285.
+- 3.1.115
+
+-------------------------------------------------------------------
 Tue Aug 16 14:16:01 UTC 2016 - lslezak@suse.cz
 
 - Remove the icon from the add-on selection popup (bsc#875201)
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -843,7 +843,8 @@ module Yast
     # @param [Any] unique ID
     # @param [String] path to directory with unpacked licenses (mandatory)
     def cache_license_acceptance_needed(id, license_dir)
-      raise "Parameter 'license_dir' must not be nil" if license_dir.nil?
+      # license_dir can be nil if there is no license present (e.g. DUDs)
+      return if license_dir.nil?
 
       license_acceptance_needed = !FileUtils.Exists("#{license_dir}/no-acceptance-needed")
       SetAcceptanceNeeded(id, license_acceptance_needed)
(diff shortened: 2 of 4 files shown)


schubi2:
I think, we are too much restricted here :-)

2016-08-18 04:26:10 <1> vagrant-sles12(22975) [Ruby] modules/ProductLicense.rb:746 No license present
2016-08-18 04:26:10 <1> vagrant-sles12(22975) [Ruby] modules/ProductLicense.rb:831 ProductLicense settings: license_dir: nil, tmpdir: nil, info_file: nil
2016-08-18 04:26:10 <3> vagrant-sles12(22975) [Ruby] yast/wfm.rb:222 Client call failed with 'Parameter 'license_dir' must not be nil' and backtrace ["/update/0

src/modules/ProductLicense.rb
```diff
@@ -843,7 +843,8 @@ def GetSourceLicenseDirectory(src_id, fallback_dir)
     # @param [Any] unique ID
     # @param [String] path to directory with unpacked licenses (mandatory)
```
lslezak: Please update the documented parameters (and fix the names).

src/modules/ProductLicense.rb
```diff
@@ -843,7 +843,8 @@ def GetSourceLicenseDirectory(src_id, fallback_dir)
...
     def cache_license_acceptance_needed(id, license_dir)
-      raise "Parameter 'license_dir' must not be nil" if license_dir.nil?
+      # license_dir can be nil if there is no license present (e.g. DUDs)
+      return if license_dir.nil?
```
lslezak: I miss a test for this. Is it possible to add an unit test? (If it is not too difficult, I have no idea about testability of this code...)

lslezak:
In general looks good, just update the doc and add a test case. :wink:

src/modules/ProductLicense.rb
lslezak: The parameter names are missing, it should be `@param [Any] id unique ID` and `@param [String,nil] license_dir path to ...`

package/yast2-packager.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 18 10:55:06 CEST 2016 - schubi@suse.de
+
+- Do not raise an exception if no license is available (no licence
+  dir at all) to accept. This is still a fix for bnc#993285.
```
mvidner: a new bug has been filed for this,  https://bugzilla.suse.com/show_bug.cgi?id=994309

test/product_license_test.rb
```diff
@@ -165,45 +165,65 @@
...
+      context "when no license exists" do
+        before do
+          # Tarball with licenses does not exist
+          allow(Yast::FileUtils).to receive(:Exists).with(/license.tar.gz/).and_return(false)
+          # Info file does not exist
+          allow(Yast::FileUtils).to receive(:Exists).with(/info.txt/).and_return(false)
+        end
+
+        it "do not blame that there is no license directory" do
+           expect(Yast::ProductLicense.AcceptanceNeeded(base_product_id)).to eq(true)
```
lslezak: Do we need to test the initial state? On the next line you set it to `false` anyway. It seems that this line is useless and should be removed...

src/modules/ProductLicense.rb
```diff
@@ -325,7 +325,7 @@ def base_product_id
     # Returns whether accepting the license manually is requied.
     #
     # @see BNC #448598
-    # @param [Any] unique ID
+    # @param [Any] id unique ID
```
lslezak: Thanks for fixing these methods as well!

test/product_license_test.rb
schubi2: No, it is needed because otherwise the needed part of code will not be reached.

lslezak: Oh, this is `base_product_id`... But still, do we need to test it?

lslezak:
LGTM

kobliha:
Good catch, thanks :)

PR #185, opened 2016-08-23
Unify license handling

jreidinger:
YaST team get report that license agreement in automatic installation differs from common installation. So we look at it and unify it. We are in quite late phase of release, so it not done as unification of code, only with adapting one dialog to look like other. Also we are after string freeze due to translations, so we use trick to use translated text from different text domain to avoid new translatable string.

At first we unify enabled Next button, which is in autoinstallation disabled until clicked to checkbox that license is agreed ( but not disabled again when unchecked checkbox ).
old_open
disabled one have greyed out next button. And now how it looks now:
new_open

And how it looks when popup appear. At first old one:
old_popup

and with new one:

new_popup

and to compare how it looks with common installation, that now it is almost same ( except that common installation integrate also language and keyboard settings ):

common

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -3,6 +3,8 @@ Tue Aug 23 11:17:36 UTC 2016 - jreidinger@suse.com
 
 - always enable next in license confirmation dialog, because even
   if not accepted, popup is shown to user (bnc#993530)
+- use same popup when license not confirmed for base product in
+  license confirmation dialog (bnc#993530)
 - 3.1.116
 
 -------------------------------------------------------------------
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug 23 11:17:36 UTC 2016 - jreidinger@suse.com
+
+- always enable next in license confirmation dialog, because even
+  if not accepted, popup is shown to user (bnc#993530)
+- 3.1.116
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


package/yast2-packager.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Tue Aug 23 11:17:36 UTC 2016 - jreidinger@suse.com
+
+- always enable next in license confirmation dialog, because even
```
lslezak: - NP:  *always enable next* might be confusing, I'd use *always enable "Next" button*
- NP2: Start with capital letter to be unified with the rest

package/yast2-packager.changes
```diff
@@ -1,4 +1,13 @@
...
+  if not accepted, popup is shown to user (bnc#993530)
+- use same popup when license not confirmed for base product in
```
lslezak: NP: definite articles: *the same popup* and *in the license confirmation dialog*

src/modules/ProductLicense.rb
```diff
@@ -471,16 +471,12 @@ def DisplayLicenseDialogWithTitle(languages, back, license_language, licenses, i
...
-      # by default
-      default_next_button_state = AcceptanceNeeded(id) ? false : true
-
       Wizard.SetContents(
         caption,
         contents,
         GetLicenseDialogHelp(),
         back,
-        default_next_button_state
+        true # always allow next button, as if not accepted, it will raise popup (bnc#993530)
```
lslezak: NP: I prefer comments on a separate line.

lslezak:
LGTM, thanks for the screenshots!

PR #186, opened 2016-08-29
Add a test to Yast::AddOnProduct.AddRepo

imobachgs:
Although it was not the real cause of the bug, a nasty error was discovered through bug [bsc#993690](https://bugzilla.suse.com/show_bug.cgi?id=993690). The problem was that `AddRepo` got some refactoring but I didn't write a unit test for it. The problem was solved in #181 and here is the unit test.

I'm not releasing a new version as the fix was already shipped and this PR does not contain any change to the code included in the package.

### DIFF
--- a/test/addon_product_test.rb
+++ b/test/addon_product_test.rb
@@ -282,4 +282,45 @@ describe Yast::AddOnProduct do
       end
     end
   end
+
+  describe "#AddRepo" do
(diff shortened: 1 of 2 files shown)


mvidner:
LGTM

imobachgs:
Thanks!

schubi2:
LGTM

PR #187, opened 2016-08-31
Fix URL when retrying to load an add-on from a DVD

imobachgs:
**This behavior is also present in SLE 12 SP1, so I would like to check if it's intended.**

While retrying to load an add-on from a CD/DVD with this dialog:

ask-for-cd

The URL gets screwed (note the duplicated `devices=/dev/sr0`):

broken-url

This patch fix this issue.

### DIFF
--- a/src/modules/AddOnProduct.rb
+++ b/src/modules/AddOnProduct.rb
@@ -2235,7 +2235,7 @@ module Yast
         end
 
         # ask for a different medium
-        current_url = AskForCD(current_url, prodname)
+        current_url = AskForCD(url, prodname)
         return nil if current_url.nil?
       end
     end
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Aug 31 14:25:19 UTC 2016 - igonzalezsosa@suse.com
+
+- Fix URLs handling when retrying to load an add-on from a CD/DVD
+  through the add_on_products.xml (related to bsc#991935).
+- 3.1.117
+
+-------------------------------------------------------------------
 Tue Aug 23 11:17:36 UTC 2016 - jreidinger@suse.com
 
 - Always enable "Next" button in license confirmation dialog,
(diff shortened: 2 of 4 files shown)


src/modules/AddOnProduct.rb
```diff
@@ -2235,7 +2235,7 @@ def add_product_from_cd(url, pth, priority, prodname, check_name)
         end
 
         # ask for a different medium
-        current_url = AskForCD(current_url, prodname)
+        current_url = AskForCD(url, prodname)
```
jreidinger: how can it be fixed when current_url equal url? Only possibility I see is that `AddRepo` modify its parameter, so maybe this is actually bug? Maybe some refactoring break behavior of method?

jreidinger: or does it get broken when assigning current_url there in next iteration of loop?

jreidinger: in general I worry currently this method is more confusing after this change and it require slightly bigger change.

imobachgs: That's the point: in the first iteration, `current_url` is set to `cd:/?devices=/dev/sr0` (and `url` and `current_url` won't point to the same string anymore).

In the second iteration, `AskForCD` will receive that value and it will append `devices=/dev/sr0`  ( because of [L1553](https://github.com/imobachgs/yast-packager/blob/5e7210c0614eb08eb51a6e105be4280c005cc92e/src/modules/AddOnProduct.rb#L1553) ) resulting in `cd:/?devices=/dev/sr0&devices=/dev/sr0`. And so on the third, fourth... iterations.

jreidinger:
ok, LGTM, making it more understandable means too much changes.

imobachgs:
@jreidinger Yes, I think so and I don't want to break anything especially now :)

lslezak:
Yes, a simpler change is preferable in the current state. However we should at least add a `FIXME` note to the [`AskForCD`](https://github.com/imobachgs/yast-packager/blob/5e7210c0614eb08eb51a6e105be4280c005cc92e/src/modules/AddOnProduct.rb#L1547) method so we can fix it later when touching the code...

PR #188, opened 2016-09-15
Reduce build time

jreidinger:
speed_up around 20%

dropped dependency on yast2-country ( but still depends on storage, so layer do not change )

rpm content verified:

```
diff -r old_rpm/usr/share/applications/YaST2/checkmedia.desktop new_rpm/usr/share/applications/YaST2/checkmedia.desktop
2d1
< X-SuSE-DocTeamID=ycc_checkmedia
25d23
< 
diff -r old_rpm/usr/share/applications/YaST2/sw_single.desktop new_rpm/usr/share/applications/YaST2/sw_single.desktop
2d1
< X-SuSE-DocTeamID=ycc_sw_single
24d22
< 
diff -r old_rpm/usr/share/applications/YaST2/sw_source.desktop new_rpm/usr/share/applications/YaST2/sw_source.desktop
2d1
< X-SuSE-DocTeamID=ycc_sw_source
24d22
< 
Only in new_rpm/usr/share/doc/packages/yast2-packager: CONTRIBUTING.md
Only in new_rpm/usr/share/doc/packages/yast2-packager: README.md
```


package/yast2-packager.spec
```diff
@@ -25,17 +25,20 @@ Source0:        %{name}-%{version}.tar.bz2
 
 Url:            https://github.com/kobliha/yast-packager
 BuildRequires:  update-desktop-files
-BuildRequires:  yast2-country-data
 BuildRequires:  yast2-devtools >= 3.1.10
-BuildRequires:  yast2-storage
-BuildRequires:  yast2-testsuite
 BuildRequires:  yast2-xml
-BuildRequires:  yast2_theme
+# needed for space calculator to parse and convert sizes
```
lslezak: NP: I'd add a `TODO` note here, depending on a full storage stack just because of parsing size strings does not look nice to me. Getting rid of this dependency would help quite a lot I guess...

kobliha: We  need to define the critical path first. Guessing what might or might not help is good, but having the data before that would be better.

kobliha:
Anyway, great work, great result :) Thanks!

package/yast2-packager.spec
jreidinger: @kobliha actually it is critical path, as it blocks moving packager to lower layer.

PR #190, opened 2016-10-04
Added yast2-sw_single-wrapper to handle rpms in KDE

ancorgs:
This should fix https://bugzilla.suse.com/show_bug.cgi?id=954143 and https://bugzilla.suse.com/show_bug.cgi?id=1002417 for Leap 42.2 and SLE-12-SP2. Since both are in beta phase, waiting for release managers approval, see comments in first bug report.

### DIFF
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,11 +17,12 @@
 
 
 Name:           yast2-packager
-Version:        3.1.117
+Version:        3.1.117.1
 Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 Source0:        %{name}-%{version}.tar.bz2
+Source1:        yast2-sw_single-wrapper
 
 Url:            https://github.com/kobliha/yast-packager
 BuildRequires:  update-desktop-files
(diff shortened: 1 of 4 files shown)


embar-:
We also should add line in .spec file (because yast2-sw_single-wrapper calls xdg-su):

Requires:       /usr/bin/xdg-su

ancorgs:
Pending of the branching and versioning issues. See thread https://lists.opensuse.org/yast-devel/2016-10/msg00003.html

package/yast2-packager.spec
```diff
@@ -101,6 +102,8 @@ This package contains the libraries and modules for software management.
 
 %install
 %yast_install
+mkdir -p %{buildroot}%{_bindir}
```
lslezak: Actually why the wrapper script cannot be part of the tarball and installed the usual way? With this approach the wrapper will not be installed when you run `rake install` from the sources. That will be confusing, if you update the file locally your changes will not take effect...

lslezak:
BTW is SP2 still relevant? Maybe `master` is enough...

teclator:
A [trello card](https://trello.com/c/tjc6qVDU/1346-leap-42-2-p5-1017696-missing-quoting-when-installing-rpm-with-yast-via-kde-dolphin-desktop-file) has been created related to this PR and also #195 as there are some controversial discussions pending and probably make sense to modify target branch to master as @lslezak commented, so just close them and plan for next sprints

It has been already discussed with @ancorgs

PR #191, opened 2016-10-05
Backport iso fix

jreidinger:


package/yast2-packager.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Oct  3 16:38:36 UTC 2016 - jreidinger@suse.com
+
+- prevent double URL encoding for ISO (bsc#954813)
+- 3.1.118
```
lslezak: I'd revert the version bump, I'd do that just before releasing the maintenance update (to avoid possible collisions later).

jreidinger: I need to release Leap 42.2 update, so as we agreed, bump version there and for critical maintenance updates use 3.1.117.1

PR #192, opened 2016-10-05
Fake mtab removal

jreidinger:


aschnell:
Sure that not even a link to /proc/self/mounts is needed?

jreidinger:
@aschnell where do you see such link creation? removed code only create fake mtab. About /proc/self/mounts I do not know, it is question to @schubi2 or lnussel

PR #194, opened 2016-10-10
Better handle disabled add-ons check box (bsc#955186)

lslezak:
Do not ask for a CD when the add-on check box is unchecked, fixes [bug 955186](https://bugzilla.suse.com/show_bug.cgi?id=955186)

- 3.2.2

### The Original Problem

See the screenshot:

- Check the *I would like to install...* check box
- Select the *CD* repository
- Uncheck the add-on check box back
- The radiobuttons are correctly disabled
- But after pressing *Next* the user is still asked for a CD medium!

add-on-disabled

### Fix

With this fix the user is not asked for a CD after unchecking the check box. The next installation dialog is displayed. Moreover adding a CD when the checkbox is enabled still works :wink: .

### DIFF
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -67,6 +67,9 @@ module Yast
     VALID_URL_SCHEMES = ["ftp", "tftp", "http", "https", "nfs",
       "nfs4", "cifs", "smb", "cd", "dvd", "iso", "dir", "file", "hd"]
 
+    # repository types which need special handling
+    SPECIAL_TYPES = [:slp, :cd, :dvd, :comm_repos, :sccrepos]
+
     def main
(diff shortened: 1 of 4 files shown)


src/modules/SourceDialogs.rb
```diff
@@ -2236,17 +2238,27 @@ def SelectHandle(key, event)
       #  TODO: disable "download" option when CD or DVD source is selected
 
       selected = UI.QueryWidget(Id(:type), :CurrentButton)
-      return :finish if [:slp, :cd, :dvd, :comm_repos, :sccrepos].include?(selected)
+      special_repo = [:slp, :cd, :dvd, :comm_repos, :sccrepos].include?(selected)
```
imobachgs: It looks like a constant.

PR #195, opened 2016-10-11
Added yast2-sw_single-wrapper to handle rpms in Plasma

ancorgs:
Replaces #190 

See
https://bugzilla.suse.com/show_bug.cgi?id=954143
https://bugzilla.suse.com/show_bug.cgi?id=1002417
https://features.opensuse.org/319376

BTW, something to fix in the future (maybe when merging to master). `desktop/yast2-packager.desktop` seems to be in the wrong location (should be `src/desktop`) and is redundant with `src/desktop/sw_single.desktop`.

### DIFF
--- a/desktop/yast2-packager.desktop
+++ b/desktop/yast2-packager.desktop
@@ -2,7 +2,7 @@
 Encoding=UTF-8
 Name=Install/Remove Software
 GenericName=Install/Remove Software
-Exec=xdg-su -c "/sbin/yast2 sw_single %F"
+Exec=/usr/bin/yast2-sw_single-wrapper %F
 Icon=yast-sw_single
 Terminal=false
 Type=Application
@@ -11,3 +11,6 @@ MimeType=application/x-rpm;application/x-redhat-package-manager;
 NotShowIn=GNOME;MATE;
 StartupNotify=true
 
+# KDE specific (does not affect the other desktop environments)
+# https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#kde-items
+InitialPreference=10
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -22,6 +22,7 @@ Release:        0
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 Source0:        %{name}-%{version}.tar.bz2
+Source1:        yast2-sw_single-wrapper
 
 Url:            https://github.com/kobliha/yast-packager
 BuildRequires:  update-desktop-files
@@ -63,6 +64,9 @@ Requires:       /usr/bin/md5sum
 # .process agent
 Requires:       yast2-core >= 2.16.35
 
+# yast2-sw_single-wrapper requires xdg-su
+Requires:       /usr/bin/xdg-su
+
 # setenv() builtin
 Conflicts:      yast2-core < 2.15.10
 
@@ -101,6 +105,8 @@ This package contains the libraries and modules for software management.
 
 %install
 %yast_install
+mkdir -p %{buildroot}%{_bindir}
+install -m 755 %{SOURCE1} %{buildroot}%{_bindir}/yast2-sw_single-wrapper
 
 %suse_update_desktop_file yast2-packager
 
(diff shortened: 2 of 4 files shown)


package/yast2-sw_single-wrapper
```diff
@@ -0,0 +1,32 @@
...
+# https://bugzilla.novell.com/show_bug.cgi?id=222757
+
+# This stripped /usr/bin/package-manager function 
+# (from libzypp (<=13.9) shipped with openSUSE 13.1 and earler)
+# to pass RPM packages to true package manager
+
+# quoted concatenation of arguments
+function quote() {
+  # formerly used 'printf %q' breaks UTF-8 strings
+  echo -n "$@" | sed 's/\([]|&;<>()$`\" \t*?#~=%[]\)/\\\1/g'
```
shundhammer: That regexp sure looks impressive, but it could do with some comments on the line(s) above it, preferably including. some examples what goes in and what comes out

package/yast2-sw_single-wrapper
```diff
@@ -0,0 +1,32 @@
+#! /bin/sh
+
+# This script helps to correctly pass RPM packages including spaces in the path
+# (and/or multiple RPM packages) to YaST2 via desktop file
+# See https://bugzilla.suse.com/show_bug.cgi?id=1002417
+
+# Originally created by http://en.opensuse.org/User:Mvidner for
+# https://bugzilla.novell.com/show_bug.cgi?id=222757
```
mvidner: The mentioned bug has a much different script. Did you adapt this one from somewhere else?

package/yast2-sw_single-wrapper
```diff
@@ -0,0 +1,32 @@
...
+
+# This stripped /usr/bin/package-manager function 
+# (from libzypp (<=13.9) shipped with openSUSE 13.1 and earler)
+# to pass RPM packages to true package manager
+
+# quoted concatenation of arguments
+function quote() {
+  # formerly used 'printf %q' breaks UTF-8 strings
```
mvidner: That sounds interesting, and would be a bug in bash. Please show me some details.

package/yast2-sw_single-wrapper
mvidner: Ah. The bug links should go to the code comments, not only to PR description, otherwise archeology becomes much harder.

desktop/yast2-packager.desktop
```diff
@@ -2,7 +2,7 @@
 Encoding=UTF-8
 Name=Install/Remove Software
 GenericName=Install/Remove Software
-Exec=xdg-su -c "/sbin/yast2 sw_single %F"
```
mvidner: I think this means that the real bug is in /sbin/yast2 and we should fix that, instead of adding yet another script which is just overly complex quoting.

ancorgs: Strange enough, I checked that calling yast2 directly (without the wrapper script) works from Thunar (XFCE file manager) and I assume it works also from Gnome's file manager. But it fails from Dolphin (KDE-Plasma-Today's_name file manager).

ancorgs:
I'm all for an alternative solution without the wrapper script. I simply went for it because it was verified to work by the reporter (and later by my) and coming from a trustworthy source (@mvidner). But I admit I'm not happy adding an extra layer either.

lslezak:
Please set the target to `master`, `merge_after_SP2_release` branch is outdated.

package/yast2-packager.spec
```diff
@@ -101,6 +105,8 @@ This package contains the libraries and modules for software management.
 
 %install
 %yast_install
+mkdir -p %{buildroot}%{_bindir}
```
lslezak: Actually why the wrapper script cannot be part of the tarball and installed the usual way? With this approach the wrapper will not be installed when you run `rake install` from the sources. That will be confusing, if you update the file locally your changes will not take effect...

teclator:
A [trello card](https://trello.com/c/tjc6qVDU/1346-leap-42-2-p5-1017696-missing-quoting-when-installing-rpm-with-yast-via-kde-dolphin-desktop-file) has been created related to this PR and also #190 as there are some controversial discussions pending and probably make sense to modify target branch to master as @lslezak commented, so just close them and plan for next sprints

It has been already discussed with @ancorgs

PR #199, opened 2016-10-14
Do not escape characters in the path string

schubi2:
I am not sure here if this is the right fix. Other suggestions are really welcome !

### DIFF
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -249,7 +249,7 @@ module Yast
         if Builtins.regexpmatch(query, "device=/dev/disk/by-id/usb-")
           Ops.set(parsed, "scheme", "usb")
 
-          @_url = URL.Build(parsed)
+          @_url = build_url(parsed)
           Builtins.y2milestone(
             "URL %1 is an USB device, changing the scheme to %2",
             URL.HidePassword(url),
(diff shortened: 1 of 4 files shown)


src/modules/SourceDialogs.rb
```diff
@@ -2713,6 +2713,19 @@ def valid_scheme?(url)
...
+  private
+
+    # Build URL from tokens
+    # @param [Hash] with keys:"scheme","http","host","port","path","query",...
+    # @return [String] url, empty string if invalid data is used to build the url.
+    def build_url(url_settings)
+      # URL.Build is escaping characters (e.g. ":") which is wrong for
+      # pathes like /dist/ibs/SUSE:/SLE-12-SP1:/GA. (bnc#966413)
+      # So we are unescaping it.
+      URLRecode.UnEscape(URL.Build(url_settings))
```
jreidinger: this looks wrong, as it also unescape other characters that are correctly escaped and cannot be in URL. I think we really need testcases for this and be sure that we return valid url, as this can seriously break installation. maybe only escaping `:` is wrong?

jreidinger: in fact it looks like bug in URL.Build as according to its source https://github.com/yast/yast-yast2/blob/5762181d62762816a73fc040362c1efb5d97deed/library/types/src/modules/URL.rb#L91 `:` should be escaped only for user and password part of URL, not for path which is your example.

schubi2:
We have a better fix in yast2 package.

PR #200, opened 2016-10-17
Update the priority for the loaded packages (bsc#498266)

lslezak:
... when starting the repository manager from the package manager

- 3.2.4

When the repository manager was started from the package management the changed repository priority was not applied on the pool. You had to restart YaST.

### DIFF
--- a/src/clients/repositories.rb
+++ b/src/clients/repositories.rb
@@ -1683,6 +1683,9 @@ module Yast
               Ops.set(sourceState, "priority", new_priority)
               Ops.set(@sourceStatesOut, global_current, sourceState)
 
+              # update the priority for the already loaded packages
+              Pkg.SourceSetPriority(sourceState["SrcId"], new_priority)
+
               # do not refresh the item in the table
               current = -1
             else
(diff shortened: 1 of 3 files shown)


src/clients/repositories.rb
```diff
@@ -1683,6 +1683,9 @@ def SummaryDialog
               Ops.set(sourceState, "priority", new_priority)
```
jreidinger: is this needed? I think it is done automatic in Pkg call or is it cache?

src/clients/repositories.rb
```diff
@@ -1683,6 +1683,9 @@ def SummaryDialog
...
               Ops.set(@sourceStatesOut, global_current, sourceState)
 
+              # update the priority for the already loaded packages
+              Pkg.SourceSetPriority(sourceState["SrcId"], new_priority)
```
jreidinger: what happen if user cancel changes in repo manager? it looks like it kept set. Maybe this call should be in store method?

lslezak: Good question, the changed value will be kept. I'll move it...

lslezak: Updated, the same problem was actually also when enabling/disabling a repository.

jreidinger:
LGTM

PR #201, opened 2016-10-24
[Review] Request from 'schubi2' @ 'yast/yast-packager/review_161024_faking_mtab'

schubi2:
  * 34f55d1 faking mtab on target system


src/modules/Mtab.rb
```diff
@@ -0,0 +1,49 @@
...
+    #
+    # Reading /etc/mtab from inst-sys, removing all /mnt headings
+    # and writing this patched mtab into target system.
+    #
+    def clone_to_target
+      log.info("Copying /etc/mtab to the target system...")
+      mtab = WFM.Read(path(".local.string"), MTABNAME)
+      # remove non-existing mount points
+      mtab_lines = mtab.split("\n")
+      mtab_lines.collect! do |mtab_line|
```
jreidinger: what about using `map!` which is more convenient?

jreidinger:
LGTM

PR #204, opened 2016-11-14
Generic solution for the YaST needed packages (bsc#885496)

lslezak:
# Description

This is a generic solution for checking the packages/patterns required by YaST itself. See the [original bug report](https://bugzilla.suse.com/show_bug.cgi?id=885496) for the bootloader module.

Currently only the bootloader module checks for unselected packages, this is a generic solution for all YaST modules. (The bootloader specific part will be removed after merging this PR.)

## Screenshots

The proposal contains an error message with list of missing packages or patterns if user deselects some needed items:

packages_proposal

After clicking the `Install` button the installation is blocked, user must resolve the problem either by selecting the packages back or by adjusting the respective YaST configuration (e.g. do not install any bootloader and disable the firewall).

packages_proposal2

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Nov 14 11:40:56 UTC 2016 - lslezak@suse.cz
+
+- Software proposal: display an error when a package or a pattern
+  required by YaST has been deselected by user, implement a generic
+  solution for bsc#885496
+- 3.2.6
+
+-------------------------------------------------------------------
 Mon Oct 24 14:54:59 CEST 2016 - schubi@suse.de
 
 - Faking /etc/mtab on target system for post RPM installation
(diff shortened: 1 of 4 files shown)


jreidinger:
well, one usability not, if would be nice to write which component need given packages ( packages should  have it )

jreidinger:
LGTM, if you tested it properly manually ;)

lslezak:
Yes, I thought about improving the usability. Unfortunately the component is only identified by an unique ID which might but might not refer to some YaST module. In reality it is:
```ruby
{
  "storage_proposal" => {:package => ["btrfsprogs", "e2fsprogs", "snapper"]},
  "DefaultDesktopPatterns" => {:pattern => ["kde", "x11", "base"]},
  "yast2-bootloader" => {:package => ["kexec-tools", "grub2", "syslinux"]}
}
```

That means you cannot easily say what the user need to do or which YaST configuration needs to be changed.

And for example storage or the desktop pattern are selected *before* entering the proposal, you need to go back to change them.

PR #205, opened 2016-11-15
Error update casp

jreidinger:


lslezak:
Um, I'm not sure if the fix is actually needed for CASP.

- CASP is a brand new product, you cannot upgrade it from any previous installation.
- Will be an upgrade possible? AFAIK there will be no upgrade for CASP, you always get a new image or install a new system.

PR #207, opened 2016-11-24
Summary in proposal

jreidinger:


lslezak:
Background: This is called from the software proposal: https://github.com/yast/yast-packager/blob/dff498280c3969db2a386e9face338af36e24a4b/src/clients/software_proposal.rb#L124

PR #209, opened 2016-12-01
Mark the default selected patterns as optional (bsc#885496)

lslezak:
Fixes broken openQA tests

- 3.2.10

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -1103,9 +1103,11 @@ module Yast
         end
       end
 
-      # FATE #302116
-      # BNC #431580
+      # FATE #302116, BNC #431580
+      # select both mandatory and optional patterns
       required_patterns = PackagesProposal.GetAllResolvables(:pattern)
+      required_patterns.concat(PackagesProposal.GetAllResolvables(:pattern, optional: true))
+
       if required_patterns != nil && required_patterns != []
         Builtins.y2milestone(
           "Patterns required by PackagesProposal: %1",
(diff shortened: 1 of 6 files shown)


src/modules/Packages.rb
```diff
@@ -1103,9 +1103,11 @@ def ComputeSystemPatternList
         end
       end
 
-      # FATE #302116
-      # BNC #431580
+      # FATE #302116, BNC #431580
+      # select both mandatory and optional patterns
       required_patterns = PackagesProposal.GetAllResolvables(:pattern)
+      required_patterns.concat(PackagesProposal.GetAllResolvables(:pattern, optional: true))
```
jreidinger: I worry this need some renaming. It is very confusing to call it required, when it include also optional.

src/modules/Packages.rb
```diff
@@ -1154,9 +1156,10 @@ def ComputeSystemPackageList
 
       # bnc #431580
       # New API for packages selected by other modules
-      packages_proposal_all_packages = PackagesProposal.GetAllResolvables(
-        :package
-      )
+      # use both mandatory and optional packages
+      packages_proposal_all_packages = PackagesProposal.GetAllResolvables(:package, optional: true)
+      packages_proposal_all_packages.concat(PackagesProposal.GetAllResolvables(:package))
```
jreidinger: below is log message about no required packages, but in fact it is no all packages, required and also optional.

src/modules/Packages.rb
```diff
@@ -1129,118 +1079,52 @@ def ComputeSystemPatternList
...
 
-      if Pkg.IsSelected("xorg-x11-Xvnc") && Linuxrc.vnc
-        install_list = Convert.convert(
-          Builtins.union(install_list, graphicPackages),
-          :from => "list",
-          :to   => "list <string>"
-        )
-      else
-        Builtins.y2milestone("Not selecting graphic packages")
+        kernel_pkgs_additional = ComputeAdditionalKernelPackages()
```
jreidinger: just idea, why not have in `ComputeAdditionalKernelPackages()` switch to detect mode and also return also `Kernel.ComputePackages` included? this way this can be simple one liner.

jreidinger:
just NP, LGTM

src/modules/Packages.rb
lslezak: Um, yes, there are still plenty improvements possible, I just gave up. This blocks openQA so it should go out rather quickly...

PR #210, opened 2016-12-08
Sle 12 sp2 casp fate 321764

schubi2:


imobachgs:
LGTM

kobliha:
@imobachgs @schubi2 Guys, I'm afraid that this will need to be done differently. You can't copy files from inst-sys to the just installed system. It's because you can't be sure whether you are then changing just those three options on the installed system. Additionally, you can install basically ANY libzypp version during the installation, which means ANY config file. This has to be changed into what the feature says: only change those three options, please.

@lslezak @jsrain Please confirm

lslezak:
Yes, copying the whole file is not the best solution, it works but it's fragile. During installation the registration step can add the Update repo with a new libzypp with updated `libzypp.conf` file.

So you should really only change the required options and keep the original values from the package for the rest.

Not copying anything from the inst-sys was one of the main requirements for the new installer (the SLE12 installer), see the [new installer requirements](https://github.com/SUSE/new-installer/blob/master/doc/requirements/SLE12-Installer-Workflow.md#other-requirements-22), esp. the `§2.2.2` note.

jsrain:
I agree, let's use the sed commands directly.

PR #211, opened 2016-12-21
Sle 12 sp2 casp fate 321764 2

schubi2:


imobachgs:
What about using `cfa` for such modifications? @jreidinger Opinion?

jreidinger:
In general I do not like this approach. Users call very long time for having ability to configure this in yast packager. I think if we start with common config file approach, we can have this feature for very low cost. But this sed single shot approach is not nice, nor extendable.

Not mentioning that there is a bug and test are very weak.

Just test this code against real file and to discover bug, try to have in documentation named option like 
```
# change onlyRequires to true if you want to install only required
solver.onlyRequires = false
```

PR #212, opened 2016-12-23
Use CFA to manage zypp.conf

imobachgs:
Work in progress to use CFA to manage zypp.conf. Problems:

* Options are always placed at the end of each section. It could be ok, but it would be better to uncomment the option if it exists. For example: `# solver.onlyRequires = false` to `solver.onlyRequires = true`.
* Empty lines are removed.
* ~~~We need to add a flag in the control file to enable this behavior so we can use the same codebase in all products.~~~ (`software/minimalistic_configuration`).


test/lib/cfa/zypp_conf_test.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  describe "#set_minimalistic!" do
+    before { config.load }
+
+    it "sets minimalistic options" do
+      config.set_minimalistic!
+      main = config.section("main")
+      expect(main["solver.onlyRequires"]).to eq("true")
+      expect(main["rpm.install.excludedocs"]).to eq("yes")
+      expect(main["multiversion"]).to be_nil
```
jreidinger: what about some write test for cfa model? and maybe some load one?

imobachgs: Yes, sure, we need it. I also wanted to address the issues mentioned in the PR description and to include them in such a test.

jreidinger:
notes from today debugging session. Playing with lense do no help, even using documented one that should work. But when I use augeas CLI tool, it will properly modify without eating anything. So I need to play a bit more with CFA what happened inside.

jreidinger:
I found that it is generic problem with all lenses. I add test and we will need to look more closely on it 
https://github.com/config-files-api/config_files_api/pull/15/commits/d4d67c212f76991018dd6ec194e4c8296f6697e2

jreidinger:
more notes from debugging I did in past and playing with it.

If I use only augeas methods to manipulate parsed tree, then it have proper content, but if I clean content and create it from scratch by augeas set, then it eats that lines.

jreidinger:
OK, I play with it even more and result is that I suggest to call it a feature :)
Problem with merging solution which will keep whitespaces is that it get really complex in configuration files which have collection of variables. So it is quite hard to do proper placing.

imobachgs:
Thanks!

PR #213, opened 2017-01-04
Use a Docker image at Travis

lslezak:
- Use the prebuilt Docker image at Travis
- Additionally a test has been fixed - the Dell system detection code was not mocked and the test used a real BIOS detection which caused a test failure on my Dell system :wink:
Fixed and improved the test to test both scenarios.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
-language: cpp
-compiler:
-    - gcc
+sudo: required
+language: ruby
+services:
+  - docker
+
 before_install:
-    # disable rvm, use system Ruby
-    - rvm reset
-    - wget https://raw.githubusercontent.com/yast/yast-devtools/master/travis-tools/travis_setup.sh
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,11 +1,11 @@
...
-    - rvm reset
-    - wget https://raw.githubusercontent.com/yast/yast-devtools/master/travis-tools/travis_setup.sh
-    - sh ./travis_setup.sh -p "rake yast2-devtools yast2-testsuite yast2 yast2-pkg-bindings yast2-storage" -g "rspec:3.3.0 yast-rake simplecov coveralls cheetah"
+  - docker build -t yast-test-image .
 script:
-    - sudo rake install
-    - COVERAGE=1 rake test:unit
+  # the "yast-travis" script is included in the base yastdevel/ruby-tw image
+  # see https://github.com/yast/docker-yast-ruby-tw/blob/master/yast-travis
+  - docker run -it -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-test-image yast-travis
```
jreidinger: I agree with mvidner that probably yast-test-image do not sound correct as it is ruby one.

PR #217, opened 2017-01-16
Killing the storage_wrapper and the target map usage.

ancorgs:
Better review by commit. Take into account that many of the lines reported as new in `SpaceCalculation.rb` are actually just re-indented (+ some small changes here and there).

This should ensure the old storage is not used during installation (at least with the default workflow).

Tested and all the error pop-ups related to Storage are gone.


src/modules/SpaceCalculation.rb
```diff
@@ -578,183 +555,149 @@ def get_partition_info
...
+        log.info "partition: mount: #{mount_name}, free: #{free_size_kib}KiB, used: #{used_kib}KiB"
+
+        mount_name = mount_name[1..-1] if remove_slash && mount_name != "/"
+
+        target_partitions << {
+          # FIXME: in the pre-storage-ng era, this used to be the short name of
+          # the type (like "ext3"). Now it's a number (libstorage-ng enum).
+          # Not easy to fix without relying in yast2-storage-ng (which contains
+          # the enum mappings).
+          "filesystem" => filesystem.type,
```
ancorgs: @shundhammer, @aschnell The problem explained in the comment is probably another symptom of something that has been pointed a couple of times before (specially by @jreidinger). The `filesystem.type` provided by libstorage should probably be something more rich than an enum. If it would be an object of `Storage::FilesystemType` (for example), we could do things like `filesystem.type.short_name`. This code is supposed to avoid relying in yast2-storage.

If you take a look to the `type?` method below, you can see more examples of the kind of things it would make sense to query to such an `FilesystemType` object.

package/yast2-packager.spec
```diff
@@ -62,6 +65,9 @@ Requires:       /usr/bin/md5sum
 # .process agent
 Requires:       yast2-core >= 2.16.35
 
+# Storage (methods to check the devicegraph)
+Requires:       libstorage-ng-ruby
```
jreidinger: well, I am not sure if it is good idea to allow other modules to use directly libstorage. In past we also do not do it to use underlaying stuff. I hope yast2-storage provide all methods needed.

src/lib/y2packager/storage_finder.rb
```diff
@@ -0,0 +1,53 @@
...
+    include Yast::Logger
+
+    # Storage object to use in order to access the current devicegraphs
+    #
+    # @return [Storage::Storage]
+    def storage_object
+      begin
+        require "y2storage"
+        Y2Storage::StorageManager.instance
+      rescue LoadError
```
jreidinger: please elaborate a bit more, why it is valid use case to not have y2storage available. It is not much obvious for me.

package/yast2-packager.spec
shundhammer: It's a library, so the very idea is to have it general enough so it's useful without any glue code.

The glue code will no doubt make it easier to use, but it should not be absolutely necessary to use any glue code.

shundhammer:
LGTM from my side, but I'd prefer some additional eyes to look at it, too (jreidinger is already on it, as I see).

src/modules/SpaceCalculation.rb
```diff
@@ -1093,7 +1036,53 @@ def btrfs_used_size(directory)
...
+
+    def filesystem_name(filesystem)
+      filesystem.blk_devices[0].name
+    end
+
+    def filesystem_mountpoint(filesystem)
+      filesystem.mountpoints[0] || ""
+    end
+
+    def type?(type_label, filesystem)
```
jreidinger: to be honest as @ancorgs pointed above I really do not see this as part of good design. I know that it is sometimes needed for old code, but for new object oriented code this approach looks horrible. I expect methods like
```
filesystem.type.ext?
filesystem.type.btrfs?
filesystem.type.to_sym # so it can be use in case statement above
```

src/lib/y2packager/storage_finder.rb
ancorgs: Neither is to me... just keeping the status quo in which there was a `storage_wrapper` to do more or less the same.

package/yast2-packager.spec
ancorgs: libstorage is already an object-oriented library which offer objects of classes like `Partition`, `Filesystem` and so on. It also models the relation between those objects, actions to go from A to B, etc.

I don't think one of yast-storage goals should be to wrap all that with another level of indirection just for the shake of preventing other parts of ruby from accessing libstorage directly. The goal is to make a better libstorage that does not need an extra layer to be useful.

On the other hand, even if we would have decided to add that extra level of indirection and wrap every libstorage class with a Y2Storage decorator (I felt tempted once in the past), the strict type checking performed by SWIG and C++ destroys the idea of duck-typing and makes class wrapping a PITA (that's why we use refinements instead of wrapper classes to decorate).

src/modules/SpaceCalculation.rb
aschnell: Use ```filesystem.get_displayname()```, renamed to ```filesystem.to_s()``` AFAIS.

The library can also provide a function to get the string for the enum, for other enums such functions already exist.

src/modules/SpaceCalculation.rb
```diff
@@ -191,29 +197,21 @@ def EvaluateFreeSpace(spare_percentage)
     end
 
     # return default ext3/4 journal size (in B) for target partition size
-    def DefaultExtJournalSize(part)
-      part = deep_copy(part)
-      if Ops.get_symbol(part, "used_fs", :unknown) == :ext2
+    def DefaultExtJournalSize(filesystem)
+      if type?(:ext2, filesystem)
```
aschnell: ```if ext2?(filesystem)```.

src/modules/SpaceCalculation.rb
aschnell: Just directly query the filesystem object you have, e.g. ```ext4?(filesystem)```.

src/modules/SpaceCalculation.rb
ancorgs: AFAICS, there is `Storage.ext4?(filesystem)`, but not `Storage.ext2?(filesystem)` or `Storage.reiserfs?(filesystem)` (for example).

On the other hand, the fact that those methods are directly floating in the `Storage` namespace kind of confirm @jreidinger's concerns about object orientation, in my opinion.

src/modules/SpaceCalculation.rb
ancorgs: Yes, `Filesystem#to_s` will make it, thanks.

ancorgs: But still, I believe that in a OOP world, it would be better to have `filesystem.type.to_s` than `filesystem.to_s` to get the filesystem type label.

ancorgs:
PR updated with fixes and the last step, leaving out the discussion about libstorage API for filesystem types.

package/yast2-packager.spec
jreidinger: ok, this approach have one disadvantage. In past it is enough to know API and interface of yast2-storage. Now users of API need to learn about yast2-storage API and also libstorage API. So broader API and even more request to have rock solid API of libstorage that are backward compatible.

src/modules/SpaceCalculation.rb
```diff
@@ -1093,7 +1038,46 @@ def btrfs_used_size(directory)
...
+    end
+
+    def filesystem_size(filesystem)
+      blk_device = filesystem.blk_devices[0]
+      # Only for local fs, NFS not supported yet in libstorage-ng
+      return 0 unless blk_device
+      blk_device.size
+    end
+
+    def filesystem_block_size(filesystem)
```
aschnell: The name of the function is confusing since it returns the block size of the underlying block device and not the block size of the filesystem. E.g. mkfs.ext4 also has a 'block-size' option.

src/modules/SpaceCalculation.rb
```diff
@@ -1093,7 +1038,46 @@ def btrfs_used_size(directory)
...
+      filesystem.blk_devices[0].region.block_size
+    end
+
+    def filesystem_name(filesystem)
```
aschnell: I also think the name of this function is confusing. I would expect a return value like "ext4".

src/modules/Packages.rb
```diff
@@ -2702,6 +2700,16 @@ def format_missing_resolvables(type, list)
         _("These items (%{type}) need to be selected to install: %{list}") % {type: type, list: list}
       end
     end
+
+    def fs_mountpoint(filesystem)
+      filesystem.mountpoints[0] || ""
+    end
+
+    def fs_name(filesystem)
```
jreidinger: is it fs name or device name? as it looks more for me like name of block device, not?

src/modules/Packages.rb
```diff
@@ -2702,6 +2700,16 @@ def format_missing_resolvables(type, list)
         _("These items (%{type}) need to be selected to install: %{list}") % {type: type, list: list}
       end
     end
+
+    def fs_mountpoint(filesystem)
+      filesystem.mountpoints[0] || ""
```
jreidinger: is it correct behavior if there is more mountpoints for given filesystem? E.g. what happen when it is btrfs with a lot of subvolumes and first one is e.g. "/var/run" and not "/" ? Or does it expect that first one is always main mount-point without subvolumes? and what if one device is mounted to multiple places? does it compute correctly free space?

src/modules/SpaceCalculation.rb
```diff
@@ -1093,7 +1038,42 @@ def btrfs_used_size(directory)
...
+      return 0 unless blk_device
+      blk_device.size
+    end
+
+    def filesystem_dev_name(filesystem)
+      filesystem.blk_devices[0].name
+    end
+
+    def filesystem_mountpoint(filesystem)
+      filesystem.mountpoints[0] || ""
```
jreidinger: btw these three methods with filesystem prefix smell a bit like it belongs to filesystem object from OOP POV. maybe using refinement or add it to API itself?

ancorgs: Those methods doing stuff less obvious (transversing intermediate objects or simplifying for the common case of just one device/mountpoint) are not wanted in the libstorage API (that's an already lost discussion :-) ). Using the yast2-storage-ng refinements is not an option here because we want to limit the dependency. It's not going to be like this in all the yast code.

src/modules/Packages.rb
ancorgs: yes, bad name. I will fix it

src/modules/Packages.rb
ancorgs: @aschnell Good question about btrfs. So far, `btrfs_fs.mountpoints[0]` returns the main mount point. But I'm not 100% sure if @aschnell is planning to change that in the future, when subvolumes module is improved.

If a volume is mounted to multiple places... it does the same it used to do before (in the target map era). The idea is to adapt the code with the minimal set of changes both in the code and in the behavior.

aschnell: The concept of "main mount point" is unclear.

With btrfs subvolumes enter the game things will change. How I cannot say yet - maybe each subvolume gets an object in the devicegraph and has no, one or several mount points.

ancorgs: Should I add a FIXME here (and a note in installer_hack.md) to revisit this again while implementing subvolumes?

src/modules/SpaceCalculation.rb
```diff
@@ -578,183 +559,147 @@ def get_partition_info
...
 
-      # installation stage - Storage:: is definitely present
-      # call Storage::GetTargetMap()
-      targets = Convert.convert(
-        WFM.call("wrapper_storage", ["GetTargetMap"]),
-        :from => "any",
-        :to   => "map <string, map>"
-      )
+      target_partitions = []
+      min_spare = 20 * 1024 * 1024 # minimum free space ( 20 MB )
```
aschnell: Don you have symbols to really write ```min_spare = 20 * MiB```. libstorage-ng provides some.

src/modules/SpaceCalculation.rb
```diff
@@ -578,183 +559,147 @@ def get_partition_info
...
+          if result == 0
+            # specific handler for btrfs
+            if used_fs == :btrfs
+              used = btrfs_used_size(tmpdir)
+              free_size -= used
+              growonly = btrfs_snapshots?(tmpdir)
+            else
+              partition = SCR.Read(path(".run.df"))
+
+              Builtins.foreach(partition) do |p|
```
aschnell: ```Buildins``` is obsolet.

src/modules/SpaceCalculation.rb
```diff
@@ -578,183 +559,147 @@ def get_partition_info
...
+                if p["name"] == tmpdir
+                  log.info "Partition: #{p}"
+                  free_size = p["free"].to_i * 1024
+                  used = p["used"].to_i * 1024
+                end
               end
             end
+
+            SCR.Execute(path(".target.bash"), "umount #{Shellwords.escape(tmpdir)}")
```
aschnell: I recommend to make the mount/umount RAII (or whatever this is called in Ruby).

aschnell:
I see a number of longer functions? Is rubocop happy with those or is it disabled?

aschnell:
LGTM

src/modules/SpaceCalculation.rb
ancorgs: This is old code. As said in the description of the pull request, some old code looks like new due to re-indentation. The scope of the PBI was to substitute the calls to the old libstorage with calls to the new one. I don't think rewriting/refactoring the old code should be done now.

src/modules/SpaceCalculation.rb
ancorgs: Same as above, this is old code I don't want to change to avoid diverging from master more than strictly needed.

src/modules/SpaceCalculation.rb
ancorgs: Same again. Old code, I don't think now is the right moment to change it (that would make harder to sync with master in the future).

src/modules/SpaceCalculation.rb
aschnell: Wouldn't it be better to avoid re-indentation then?

ancorgs: The point is that with the new libstorage we need way less nesting to reach the meaningful information in the data structure (4 fewer levels!). When I started with this, I consider it to not be important because I expected a lot of old code to die... but it didn't happen.

I will change this commit to keep the original indentation, even if that breaks our coding standard about indentation. I will add a comment to fix the indentation after TheFinalBigMerge.

ancorgs:
Updated the confusing commit (indentation problem) via "git push -f". Please re-approve after reviewing.

aschnell:
LGTM

PR #220, opened 2017-01-24
do not check content file on media if there is none in the ramdisk (fate#322276)

jsrain:


src/modules/Packages.rb
```diff
@@ -1269,9 +1269,15 @@ def CheckContentFile(source)
         )
         return true
       end
+      ramdisk = Convert.to_string(SCR.Read(path(".target.string"), "/content"))
+      if ramdisk == nil
```
lslezak: Maybe a simple `if !File.exists?("/content")` would be better here. This legacy code is a bit cryptic...

src/modules/Packages.rb
```diff
@@ -1269,9 +1269,15 @@ def CheckContentFile(source)
...
+        Builtins.y2milestone(
+          "Ramdisk does not contain content file, not checking the file on media"
```
lslezak: NP: `/content` file would be better

PR #221, opened 2017-01-25
Sumit to 'casp10' target

teclator:
I think that is safer to use the casp10 target and just modify it by hand if you want to use other target.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,6 @@
 require "yast/rake"
 
-Yast::Tasks.submit_to(ENV["YAST_SUBMIT"] ? ENV["YAST_SUBMIT"].to_sym : :sle12sp2)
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -1,6 +1,6 @@
 require "yast/rake"
 
-Yast::Tasks.submit_to(ENV["YAST_SUBMIT"] ? ENV["YAST_SUBMIT"].to_sym : :sle12sp2)
```
lslezak: I'd still keep `YAST_SUBMIT` handling in place, it might be useful in some cases...

imobachgs: Then we should do this in other CASP projects to be consistent.

teclator: I agree with @imobachgs, I will maintain it by now.

PR #225, opened 2017-01-30
Do not allow changing the plugin services (bsc#1021117)

lslezak:
- This is an improvement for the fix for bug [bsc#1021117](https://bugzilla.suse.com/show_bug.cgi?id=1021117) in pkg-bindings (proposal: https://github.com/yast/yast-pkg-bindings/pull/64) which just avoids a hard error at save
- This change prevents from modifying any libzypp plugin service and its repositories. They cannot be modified because there is no file where the changes could be saved, the repository metadata are dynamically generated by a script (see example in the https://github.com/yast/yast-pkg-bindings/pull/64 pull request)
- See more details at https://doc.opensuse.org/projects/libzypp/SLE12SP2/zypp-plugins.html#plugin-services and https://doc.opensuse.org/projects/libzypp/SLE12SP2/zypp-services.html
- The change is big because I have split a 900 (!) lines long method into smaller parts. Each method checks at the beginning whether the edited service or repository is of the `plugin` type. If yes a message is displayed and the change is not allowed.
- That means you do not have to check the complete diff, just the added checks.
- Tested manually
- No unit tests as the code is in a client and it is UI related, making it testable would require a major rewrite... :grimacing:
- The code would need even more refactoring, but that is out of scope for an L3 bug fix enhancement...

### DIFF
--- a/src/clients/repositories.rb
+++ b/src/clients/repositories.rb
@@ -12,6 +12,8 @@
 #
 module Yast
   class RepositoriesClient < Client
+    include Yast::Logger
+
     NO_SERVICE = :no_service
     NO_SERVICE_ITEM = :no_service_item
 
@@ -567,7 +569,7 @@ module Yast
       Builtins.foreach(deleted_services) do |_alias|
         Builtins.y2milestone("Removing service %1", _alias)
         success = success && Pkg.ServiceDelete(_alias)
-      end 
+      end
 
 
       Builtins.y2milestone("New service config: %1", @serviceStatesOut)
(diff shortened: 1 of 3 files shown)


src/clients/repositories.rb
```diff
@@ -1943,6 +1596,433 @@ def StartInstSource
...
+      # refresh the value in the table
+      new_keep = UI.QueryWidget(Id(:keeppackages), :Value)
+      log.info("New keep packages option: #{new_keep}")
+
+      sourceState["keeppackages"] = new_keep
+      @sourceStatesOut[global_current] = sourceState
+    end
+
+    # Handle the "Edit" button in the repository view
+    def repo_replace_handler(id, sourceState, global_current)
```
imobachgs: np: What about adding some documentation for parameters?

PR #227, opened 2017-02-01
Merge sp2

jreidinger:


package/yast2-packager.changes
```diff
@@ -1,32 +1,36 @@
...
+- Fix editing of iso:/ without triple slash that previously caused
+  lost of first element of path
+- 3.1.121
+
+-------------------------------------------------------------------
+Mon Jan 30 10:57:14 UTC 2017 - igonzalezsosa@suse.com
+
+- Packages module is able to perform a package selection proposal
+  during system upgrade (bsc#1009834)
+- 3.1.120
```
ancorgs: I don't get why are we rewriting history in this merge. This is a merge from into CASP. Before the merge, CASP's changelog looked like:
```
- 3.1.119.4 -> A change that is both in SP2 and CASP
- 3.1.119.3 -> Change added for CASP
- 3.1.119.2 -> Change added for CASP
- 3.1.119.1 -> Change added for CASP
- 3.1.119 -> Here we are in sync with SP2
```
And after the merge, the changelog will look like
```
- 3.1.121.1 -> A collection of previous CASP changes from all the versions that disappeared
- 3.1.121 -> A new change that is in both SP2 and CASP
- 3.1.120 -> The change (in both SP2 and CASP) that was previously referred as 3.1.119.4
- 3.1.119 -> Here we are in sync with SP2
```
I don't think rewriting the history like that helps, to be honest.

ancorgs: In the past (before CASP), when we had a similar situation with changes going to both branches we used to simply do this.
```
- 3.1.121.1 -> A new change that is in both SP2 and CASP
- 3.1.119.4 -> A change that is both in SP2 and CASP
- 3.1.119.3 -> Change added for CASP
- 3.1.119.2 -> Change added for CASP
- 3.1.119.1 -> Change added for CASP
- 3.1.119 -> Here we are in sync with SP2
```
That's more clear to me.

jreidinger: but it will change history of SP2 when we merge it. Problem is that from customer POV, if he have version 3.1.120 and then get update he check only changes in 3.1.121 and do not check history what is different

PR #228, opened 2017-02-02
More iso fixes

jreidinger:
tested manually and also added a lot of unit tests that discover various culprits founds during this manual testing

### DIFF
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -387,6 +387,8 @@ module Yast
       param_url = params.delete("url") || ""
       processed = URI(URI.encode(param_url))
       processed.scheme = "iso" if processed.scheme.downcase == "dir"
+      # we need to construct path from more potential sources, as url can look like
+      # `iso:/subdir?iso=test.iso&path=dir%3A%2Finstall` resulting in path "/install/subdir/test.iso"
       processed.path = File.join(processed.path || "", uri.path, params.delete("iso") || "")
       processed.query = URI.encode_www_form(params) unless params.empty?
 
--- a/src/modules/SourceDialogs.rb
+++ b/src/modules/SourceDialogs.rb
@@ -375,12 +375,19 @@ module Yast
     def PreprocessISOURL(url)
       log.info "Preprocessing ISO URL: #{URL.HidePassword(url)}"
 
-      uri = (url == "iso://") ? URI("iso:/") : URI(url)
+      # empty iso is used when we adding new url
+      return "" if url == "iso://"
+      uri = URI(url)
       query = uri.query || ""
-      params = URI.decode_www_form(query).to_h
 
-      processed = URI(params.delete("url") || "")
-      processed.path = File.join(processed.path || "", params.delete("iso") || "")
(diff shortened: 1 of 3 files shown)


test/source_dialogs_test.rb
```diff
@@ -7,7 +7,33 @@
 describe Yast::SourceDialogs do
   subject { Yast::SourceDialogs }
 
+  ZYPP_URL_MAPPINGS = [
```
lslezak: A mapping maps one value to another one, this is a plain list. I'd prefer a better name, like `ZYPP_VALID_URLS` or something like that.

test/source_dialogs_test.rb
```diff
@@ -47,6 +73,27 @@
...
+    end
+
+    it "handles properly escaped spaces" do
+      converted = "iso:/install/Duomenys%20600%20GB/openSUSE-13.2-DVD-x86_64.iso"
+      url = "iso:///?iso=openSUSE-13.2-DVD-x86_64.iso&url=dir%3A%2Finstall%2FDuomenys%20600%20GB"
+
+      expect(subject.PreprocessISOURL(url)).to eq(converted)
+    end
+
+    it "handles empty iso uri" do
```
lslezak: NP: Could you add a comment why we need to handle this? It's clear from the diff, but few years later it might be quite difficult to find out...

src/modules/SourceDialogs.rb
```diff
@@ -375,12 +375,19 @@ def IsISOURL(url)
...
 
-      processed = URI(params.delete("url") || "")
-      processed.path = File.join(processed.path || "", params.delete("iso") || "")
+      # libzypp do not use web encoding as in https://www.w3.org/TR/html5/forms.html#url-encoded-form-data
+      # but percentage enconding only. For more details see (bsc#954813#c20)
+      params = URI.decode_www_form(query.gsub(/%20/, "+")).to_h
+
+      param_url = params.delete("url") || ""
+      processed = URI(URI.encode(param_url))
+      processed.scheme = "iso" if processed.scheme.downcase == "dir"
```
lslezak: What about the other scheme like `file`, we do not need to handle them? Please add a comment if needed...

src/modules/SourceDialogs.rb
```diff
@@ -375,12 +375,19 @@ def IsISOURL(url)
...
+      processed.path = File.join(processed.path || "", uri.path, params.delete("iso") || "")
```
lslezak: Also mention why we need to join the paths.

test/source_dialogs_test.rb
jreidinger: true, originally it was mapping and then I realize that pre and post process is iso scheme only

test/source_dialogs_test.rb
jreidinger: ok

src/modules/SourceDialogs.rb
jreidinger: well, in Post we do reverse change, so just for sack of completeness. In fact later in dialog it is not visible, as it use only path part of url.

src/modules/SourceDialogs.rb
jreidinger: ok

PR #230, opened 2017-02-08
[WIP] Ask when the added repository is unsigned (bsc#1009127)

lslezak:


teclator:
@lslezak what is the state of this PR? still valid once the original behavior was restored? it is waiting for something?

lslezak:
I think we can close it as libzypp restored back the original behavior, see https://bugzilla.suse.com/show_bug.cgi?id=1009127#c18.

PR #232, opened 2017-02-20
- adapted to libstorage-ng change

aschnell:


aschnell:
Failure of checks is caused by checks using previous version on libstorage-ng.

shundhammer:
LGTM

PR #235, opened 2017-03-07
automatic product selection

schubi2:


src/modules/Packages.rb
```diff
@@ -2066,32 +2066,50 @@ def SelectProduct
...
       # no product selected -> select them all
       ret = true
-      if Builtins.size(selected_products) == 0
-        Builtins.y2milestone("No product selected so far...")
-        Builtins.foreach(products) do |p|
-          product_name = Ops.get_string(p, "name", "")
-          if !Builtins.regexpmatch(product_name, "-migration$")
-            Builtins.y2milestone("Selecting product %1", product_name)
-            ret = Pkg.ResolvableInstall(product_name, :product) && ret
+      unless products.find_index { |p| p["status"] == :selected }
```
imobachgs: Why not use `any?` (or even `none?` if you want to replace `unless` by `if`) instead of `find_index`? After all, you're not interested in the index value and `any?` express better your intention.

src/modules/Packages.rb
```diff
@@ -2066,32 +2066,50 @@ def SelectProduct
...
-            Builtins.y2milestone("Ignoring migration product: %1", product_name)
+            log.info("Selecting product #{product_name}")
+            selected_products << product_name
+          end
+        end
+
+        # Due selecting all available products there can be products which
+        # are conflicting due e.g. renaming eachother. These conflicts
+        # can be solved automatically.
+        selected_products.select! do |old_product|
```
imobachgs: np: You could extract this part to a separate method to make this one shorter and easier to read.

src/modules/Packages.rb
```diff
@@ -2066,32 +2066,50 @@ def SelectProduct
...
+          renamed = false
```
imobachgs: I think you could get rid of the `renamed` variable if you use `any?` instead of `each?` in the following line.

src/modules/Packages.rb
```diff
@@ -2066,32 +2066,50 @@ def SelectProduct
...
+            if AddOnProduct.renamed?(old_product, new_product)
+              log.info("Product #{old_product} will be renamed by or is included in #{new_product}")
+              renamed = true
+            end
           end
+          !renamed
+        end
+
+        selected_products.each do |name|
+          ret = Pkg.ResolvableInstall(name, :product) && ret
```
imobachgs: Should you log an error if some product cannot be resolved?

src/modules/Packages.rb
lslezak: You can simplify it to:

```ruby
selected_products.reject! do |old_product|
  selected_products.any? do |new_product|
    renamed = AddOnProduct.renamed?(old_product, new_product)
    log.info "..." if renamed
    renamed
  end
end
```

src/modules/Packages.rb
```diff
@@ -2066,32 +2066,50 @@ def SelectProduct
...
+          selected_products.each do |new_product|
+            if AddOnProduct.renamed?(old_product, new_product)
+              log.info("Product #{old_product} will be renamed by or is included in #{new_product}")
+              renamed = true
+            end
           end
+          !renamed
+        end
+
+        selected_products.each do |name|
```
lslezak: `all?` is short circuit: 

```ruby
selected_products.all? { |name| Pkg.ResolvableInstall(name, :product) }
```

Then you probably do not need the `ret` variable at all..

src/modules/AddOnProduct.rb
```diff
@@ -145,7 +145,9 @@ def main
         "sle-hae"       => [ "sle-ha" ],
         "SUSE_SLES_SAP" => [ "SLES_SAP" ],
         # SMT is now integrated into the base SLES
-        "sle-smt"       => [ "SLES" ]
+        "sle-smt"       => [ "SLES" ],
+        # SLES_SAP includes SLES. So SLES is not needed anymore
+        "SLES"          => [ "SLES_SAP" ]
```
lslezak: Um, this looks like a very interesting idea, but unfortunately this will probably not work correctly as it has some side effects.

This mapping is used when evaluating product upgrades, see https://github.com/yast/yast-packager/blob/d3c1f7e566612e507dc835f36dce879541bd44ec/src/modules/Packages.rb#L2655

In this case the SLES12 -> SLES4SAP12 upgrade would be considered OK, while originally it would report an error (a product removal) and a new product installation. And the upgrade would be blocked until user manually approved this (by manually selecting the product to install). With this change it would be considered OK.

IMHO every product change should be reported as a error to avoid changing the system by mistake (e.g. a typo in the URL). Although SLES4SAP12 is compatible with SLES12 I still think the user should explicitly confirm the upgrade to avoid surprises later.

src/modules/Packages.rb
lslezak: It's OK, this only marks the product to install, it cannot fail. It may fail later when running the solver (`Pkg.Solve`).

It can only fail when the product with name `name` cannot be found, but as we query the available products first this also cannot happen.

src/modules/Packages.rb
```diff
@@ -2066,32 +2066,50 @@ def SelectProduct
...
-        Builtins.y2milestone("Second stage - skipping product selection")
+        log.info("Second stage - skipping product selection")
         return true
       end
 
       products = Pkg.ResolvableProperties("", :product, "")
 
-      if Builtins.size(products) == 0
-        Builtins.y2milestone("No product found on media")
+      if products.empty?
```
schubi2: It can be :-)

src/modules/Packages.rb
schubi2: cool !

src/modules/AddOnProduct.rb
lslezak: OK, after thinking a bit more about it it turns out that it should actually work. The code should not preselect any conflicting products in first place.

And evaluating the products dependencies would be a bit tricky, e.g. you have product A and product B which conflicts with A. Which one should we select? A or B?

That means we cannot provide a generic solution which would work in all cases... :frowning_face: 
We might design it so it works in the SLES4SAP case, but it might not work in other cases. And the result is the same as if would use a simple hardcoded solution.

Anyway, I still think that for this purpose we should create a separate mapping, i.e. we have `@product_renames` so create `@product_conflicts` and use it for selecting products.

This would be a hardcoded list, which I do not like :frowning_face:, but I guess we cannot provide a better and still simple generic solution for all cases...

The rationale is that while `@product_renames` contains the very same products where only with  internal ID has been changed, for the conflicting products it is a different case. These products are really different, there is no connection between them.

src/modules/Packages.rb
```diff
@@ -29,6 +29,13 @@ class PackagesClass < Module
     # Minimum set of packages required for installation with remote X11 server
     REMOTE_X11_BASE_TAGS = [ "xorg-x11-server", "xorg-x11-fonts", "icewm" ]
 
+    # Some products are already be "included" in other products. So they MUST
+    # not be installed anymore because they are conflicting to the other one.
+    PRODUCT_CONFLICTS = {
+      # SLES_SAP includes SLES. So SLES is not needed anymore
```
lslezak: I'd add more details about the conflicts here, e.g.:

```ruby
# SLES_SAP contains "Conflicts: sles-release"
# see https://build.suse.de/package/view_file/SUSE:SLE-12-SP2:GA/_product/SLES_SAP-release.spec?expand=1
```

So everybody can see how the conflict is defined and it is possible to check it in the future whether the conflict is still there.

src/modules/Packages.rb
```diff
@@ -2795,6 +2797,14 @@ def kept_products(products)
     def has_window_manager?
       Pkg.IsSelected("windowmanager") || Pkg.IsProvided("windowmanager")
     end
+
+    # Checking if two products are conflicting with each other.
```
lslezak: Um, it's not exactly `with each other`. It checks whether `product1` conflicts with `product2`. It is irrelevant whether `product2` conflicts with `product1` or not.

The check is only one way.

src/modules/Packages.rb
```diff
@@ -29,6 +29,13 @@ class PackagesClass < Module
     # Minimum set of packages required for installation with remote X11 server
     REMOTE_X11_BASE_TAGS = [ "xorg-x11-server", "xorg-x11-fonts", "icewm" ]
 
+    # Some products are already be "included" in other products. So they MUST
+    # not be installed anymore because they are conflicting to the other one.
+    PRODUCT_CONFLICTS = {
+      # SLES_SAP includes SLES. So SLES is not needed anymore
+      "SLES" => [ "SLES_SAP" ]
```
lslezak: Maybe we should change the mapping direction? Because in reality it is *SLES_SAP* which conflicts with *SLES*.

The conflicts goes from *SLES_SAP* so logically it would be better to have `"SLES_SAP" => [ "SLES" ]` mapping. (And update the code respectively.)

src/modules/Packages.rb
```diff
@@ -2072,19 +2079,19 @@ def SelectProduct
...
       # no product selected -> select them all
       ret = true
-      unless products.find_index { |p| p["status"] == :selected }
+      unless products.any? { |p| p["status"] == :selected }
         log.info("No product selected so far...")
         selected_products = []
         products.each do |p|
           product_name = p["name"] || ""
-          if Builtins.regexpmatch(product_name, "-migration$")
+          if product_name.match(/-migration$/)
```
lslezak: Just a note: in Ruby the end of string matcher is `\z`, `$` means end of line (`\n`). But as the product name cannot be multi-line it does not matter in this case...

(The Ruby way is a bit complicated, see http://ruby-doc.com/docs/ProgrammingRuby/html/language.html#UL for details.)

test/packages_test.rb
```diff
@@ -1016,4 +1016,85 @@ def product(properties = {})
...
+    before do
+      allow(subject).to receive(:Initialize).with(true)
+    end
+
+    context "when it is called in continue stage" do
+      before do
+        allow(Yast::Stage).to receive(:cont).and_return(true)
+      end
+
+      it "do not install any product" do
```
lslezak: Bad English: Use `does not` here. If in doubts it helps to read it as an English sentence: `it does not install ....`

test/packages_test.rb
```diff
@@ -1016,4 +1016,85 @@ def product(properties = {})
...
+        expect(Yast::Pkg).not_to receive(:ResolvableInstall)
+        subject.SelectProduct
+      end
+    end
+
+    context "when it is called in installation stage" do
```
lslezak: NP: That would mean `Yast::Stage.initial` call, but according to the mock below it actually is `when called outside the second stage`.

test/packages_test.rb
```diff
@@ -1016,4 +1016,85 @@ def product(properties = {})
...
+      end
+
+      context "when one or more products have already been selected for installation" do
+        before do
+          allow(Yast::Pkg).to receive(:ResolvableProperties).and_return(
+            [product("name" => "p1", "status" => :selected), product("name" => "p2")]
+          )
+        end
+
+        it "do not install any additional product" do
```
lslezak: Again, `does not`.

test/packages_test.rb
```diff
@@ -1016,4 +1016,85 @@ def product(properties = {})
...
+          subject.SelectProduct
+        end
+      end
+
+      context "when no installable product has been found" do
+        before do
+          allow(Yast::Pkg).to receive(:ResolvableProperties).and_return([])
+        end
+
+        it "do not install any product" do
```
lslezak: Again, `does not`.

test/packages_test.rb
```diff
@@ -1016,4 +1016,85 @@ def product(properties = {})
...
+        before do
+          allow(Yast::Pkg).to receive(:ResolvableProperties).and_return(
+            [ product("name" => "installed_product", "status" => :installed),
+              product("name" => "p1"),
+              product("name" => "p2")
+            ]
+          )
+        end
+
+        it "do not install already installed products" do
```
lslezak: Again, `does not`.

test/packages_test.rb
```diff
@@ -1016,4 +1016,85 @@ def product(properties = {})
...
+      context "when conflicting products should be installed (SLES/SLES_SAP)" do
+        before do
+          allow(Yast::Pkg).to receive(:ResolvableProperties).and_return(
+            [ product("name" => "SLES"),
+              product("name" => "SLES_SAP")
+            ]
+          )
+        end
+
+        it "do not install SLES product" do
```
lslezak: Again, `does not`.

src/modules/Packages.rb
lslezak: Actually as we dropped the 2nd stage we should update the code from `return true if Stage.cont` to `return true unless Stage.initial`.

package/yast2-packager.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Mar  7 09:37:15 CET 2017 - schubi@suse.de
+
+- Product selection: Do not reinstall already installed products
+  or products which are already included in another product.
```
lslezak: Maybe I'd add a note about the real use case as this generic description does not say whether a customer need the fix or not. Something like `Fixes SLES4SAP autoupgrade with SUSE Manager repositories.`

PR #237, opened 2017-03-16
Uncommenting

jreidinger:


test/lib/cfa/zypp_conf_test.rb
```diff
@@ -42,7 +42,15 @@
...
 
+    it "do nothing if no option is not modified" do
+      config.save
+      expect(File.read(zypp_conf_path)).to eq(File.read(ZYPP_CONF_EXAMPLE))
+    end
+
     it "modifies the file accordingly to given options" do
+      # FIXME: now expected file include also whitespace changes caused
+      # by https://github.com/hercules-team/augeas/issues/450
+      # modify it when augeas is fixed
```
mvidner: What about doing some basic filtering and comparing the result: non-empty lines, non-comment lines only.

jreidinger: well, then we are in risk that we overlook bigger changes, like eating empty lines like it is before. Also as now zypp.conf is quite heavily commented out, there is a risk we do not see if key is properly placed.

mvidner: Hm, OK

PR #240, opened 2017-03-24
Do not ignore solver problems after touching the partitioning

mvidner:
[bsc#1029306](https://bugzilla.suse.com/show_bug.cgi?id=1029306)
https://trello.com/c/HNgqut0H


src/lib/packager/clients/software_proposal.rb
```diff
@@ -0,0 +1,169 @@
...
+# To contact SUSE about this file by physical or electronic mail, you may find
+# current contact information at www.suse.com.
+# ------------------------------------------------------------------------------
+
+require "installation/proposal_client"
+
+module Yast
+  class SoftwareProposalClient < ::Installation::ProposalClient
+    def initialize
+      Yast.import "Pkg"
```
jreidinger: is there reason to not have imports together?

src/lib/packager/clients/software_proposal.rb
```diff
@@ -0,0 +1,169 @@
...
+      Yast.import "Packages"
+      Yast.import "Language"
+      Yast.import "Installation"
+    end
+
+    def make_proposal(flags)
+      @force_reset = flags.fetch("force_reset", false)
+
+      @language_changed = adjust_locales
+      @language_changed ||= flags.fetch("language_changed", false)
```
jreidinger: well, it is usually discouraged to use `||=` for booleans, as it is not easy to recognize if it is mistake or intention.

src/lib/packager/clients/software_proposal.rb
```diff
@@ -0,0 +1,169 @@
...
+
+      # if only partitioning has been changed just return the current state,
+      # don't reset to default (bnc#450786, bnc#371875)
+      if partitioning_changed? && !@language_changed && !@force_reset && !Packages.PackagesProposalChanged
+        @ret = Packages.Summary([ :product, :pattern, :selection, :size, :desktop ], false);
```
jreidinger: is there reason to have `@ret` as class variable? it is a bit confusing.

PR #245, opened 2017-04-20
[RFC] escape backslashes in url (bsc#1032506)

jreidinger:


src/modules/Packages.rb
```diff
@@ -1600,6 +1600,10 @@ def Initialize_BaseInit(show_popup, base_url, log_url)
         base_url.value = InstURL.installInf2Url("")
       end
 
+      # '\' is not reserved nor unreserved according to rfc https://tools.ietf.org/html/rfc3986#section-2.1
+      # but ruby URI does not handle it. So to make life easier encode it (bsc#1032506)
+      base_url.value.gsub!(/\\/, "%5C")
```
lslezak: NP: If you do not need any regexp feature you can use simple `String` parameter, e.g. `gsub!("\\", "%5C")`.

PR #247, opened 2017-05-19
use only y2storage classes

wfeldt:
I feel we are maybe mocking a bit too much in space_calculation_test.rb. It doesn't rely on any
(real) y2storage class features at all...

### DIFF
--- a/test/space_calculation_test.rb
+++ b/test/space_calculation_test.rb
@@ -2,9 +2,7 @@
 
 require_relative "./test_helper"
 require 'yaml'
-require "yast"
-require "storage"
-require "y2storage"
+require 'y2storage'
 
 Yast.import 'Stage'
 Yast.import 'Mode'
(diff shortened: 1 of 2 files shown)


test/space_calculation_test.rb
```diff
@@ -26,17 +23,12 @@ def expect_to_execute(command)
...
 
-def fs_type(name)
-  Storage.const_get(:"FsType_#{name.to_s.upcase}")
-end
-
 def filesystem(size_k: 0, block_size: 4096, type: :ext2, tune_options: "")
-  region = double("Storage::Region", block_size: block_size)
-  part = double("Storage::Partition", name: "/dev/sda1", region: region, size: size_k * 1024)
+  region = double("Y2Storage::Region", block_size: block_size)
+  blk_device = double("Y2Storage::BlkDevice", name: "/dev/sda1", region: region, size: size_k * 1024)
```
ancorgs: This looks wrong. `Y2Storage::Region#block_size` and `Y2Storage::BlkDevice#size` return `Y2Storage::DiskSize` objects, not integers. So the mocks are wrong.

PR #248, opened 2017-05-23
Fix breakage introduced by previous PR

ancorgs:
#247 broke openQA. I hope this will fix it.

The first commit adjust the sizes in the test because the tests were failing in OBS (and also locally for me `rake test:unit` or ' rake osc:build`). Not sure how they managed to work before and even to pass travis recently.

The second test should fix the breakage in openQA. Verified locally.


teclator:
LGTM

PR #255, opened 2017-07-26
Remove content

jreidinger:


src/modules/AddOnProduct.rb
```diff
@@ -1196,7 +955,7 @@ def Integrate(srcid)
       end
 
       # FATE #302398: PATTERNS keyword in content file
-      HandleProductPATTERNS(srcid)
+      # changed for software section in installation.xml
```
lslezak: `# or 'defaultpattern()' product Provides (FATE#320199)`

src/modules/Packages.rb
```diff
@@ -1814,24 +1728,19 @@ def Initialize_StageInitial(show_popup, base_url, log_url)
...
-          if !Popup.ContinueCancel(
-              # message popup, %1 is product name
-              Builtins.sformat(_("Insert %1 CD 1"), label)
-            )
-            @init_error = Builtins.sformat(_("%1 CD 1 not found"), label)
-            @init_in_progress = false
-            return
-          end
+
+        # bug #159754, release the mounted CD
```
lslezak: You have to remove the whole block, the condition checks whether the content file downloaded from medium is the same as the downloaded by linuxrc. If they are not the same the medium has been changed in the mean time.

Othwerwise the installation repository will be always removed and user repeatedly asked for CD1.

We do not have an equivalent check and I don't think it is a problem removing it.

package/yast2-packager.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jul 26 14:18:31 UTC 2017 - jreidinger@suse.com
+
+- drop reading content file (FATE#322386)
```
lslezak: I'd put quotes here as the file is really named "content".

PR #256, opened 2017-07-27
eula acceptence moved to add-on procuct client

schubi2:
It is possible to put add-on products on the same medium as the base product. In that case the EULA acceptance and adding process of the add-on-product will be made in a very early installation stage (before AutoYaST has been initialized). Now in case of an AutoYaST installation this will be done later in the general add-on YaST module which is configured by the AutoYaST configuration file. So the customer has to add it  in order to handle the EULA acceptance.:
`
  <add-on>

    <add_on_products config:type="list">

      <listentry>

       <media_url><![CDATA[cd:///oes]]></media_url>

       <product>Open_Enterprise_Server</product>

       <alias>oes</alias>

       <product_dir>/</product_dir>

       <confirm_license config:type="boolean">false</confirm_license>

      </listentry>

    </add_on_products>

  </add-on>
`

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Jul 27 11:12:01 CEST 2017 - schubi@suse.de
 
-- AutoYaST: Configuring EULA acception of add-on products has
+- AutoYaST: Configuring EULA acceptance of add-on products has
   not worked if the add-on product is on the medium of the base
   product too. (bnc#1032523).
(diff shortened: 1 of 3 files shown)


package/yast2-packager.changes
```diff
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Jul 27 11:12:01 CEST 2017 - schubi@suse.de
+
+- AutoYaST: Configuring EULA acception of add-on products has
```
imobachgs: acception -> acceptance ?

PR #259, opened 2017-08-04
Caasp merge

jreidinger:


jreidinger:
@lslezak what repo you use for SP2 docker? it probably missing newer CFA

teclator:
The repo is the OBS one, so @lslezak already fixed an issue syncing them, and after some rebuild and gitignore update Travis is not complaining anymore.

package/yast2-packager.changes
```diff
@@ -22,6 +22,19 @@ Tue Mar  7 09:37:15 CET 2017 - schubi@suse.de
 - 3.1.122
 
 -------------------------------------------------------------------
+Fri Mar 24 13:53:40 UTC 2017 - mvidner@suse.com
```
lslezak: Is it OK to change the history for a maintained package?

teclator: It is the CASP history, so we have the SP2 history and the CASP history, which one should be the one selected if for example the next maintenance update is for CASP?

lslezak: I see, good point. So I'm fine with the change.

PR #261, opened 2017-08-14
Merge storage-ng branch

ancorgs:
We should probably delay the merge of this one until https://build.opensuse.org/request/show/516760 finds its way to Factory. Otherwise, that fix will be trapped in the storage-ng trap for quite some time.


ancorgs:
According to @kobliha, attracting https://build.opensuse.org/request/show/516760 into the storage-ng vortex is not that bad. Being Tumbleweed, a fix managing addons can wait a couple of months. So please review so we can merge this and move on.

PR #263, opened 2017-08-18
Merge product class

imobachgs:
We have two different `Product` classes (well, three if we count the YaST module). So this PR is about:

* Unifying installation/product and packages/product into a `Y2Packager::Product`.
* `Repository` (yast2), `ProductReader` and `ProductSorter` (yast2-installation) are also moved to `Y2Packager`.
* `Product` class get methods to handle licenses.


src/lib/y2packager/product.rb
```diff
@@ -0,0 +1,144 @@
...
+      Yast::Pkg.ResolvableProperties(name, :product, "").any? do |res|
+        res["status"] == :selected
+      end
+    end
+
+    # select the product to install
+    # @return [Boolean] true if the product has been sucessfully selected
+    def select
+      log.info "Selecting product #{name} to install"
+      Yast::Pkg.ResolvableInstall(name, :product, "")
```
lslezak: This is a bit tricky as the `Y2Packager::Product` object has also the `arch`, `version` and `vendor` attributes but this method ignores them.

So the question is if this API should be bound to the very specific libzypp product (as specified in the object attributes) or can be generic. In the original case we do not care about the version, we leave the responsibility on libzypp to select the highest version.

The API might be confusing, `Y2Packager::Product.new(name: "SLES", version: "15").select` might actually select a different version to install.

If this is OK (as it makes the implementation much easier) is should be at least documented in the method comment.

src/lib/y2packager/product.rb
```diff
@@ -0,0 +1,169 @@
...
+      Yast::Pkg.PrdNeedToAcceptLicense(name)
+    end
+
+    # Confirm the license for the product
+    def confirm_license
+      Yast::Pkg.PrdMarkLicenseConfirmed(name)
+    end
+
+    # Unconfirm the license for the product
+    def unconfirm_license
```
imobachgs: You're right. It's already addressed on #267. I followed your suggestion and replaced confirm/unconfirm with a single method which receives a boolean.

lslezak:
Um, what about that "uncorfirm" term (bad English) used several times? I know that changing the pkg-bindings API is not easy but at least the test cases could be rephrased....

src/lib/y2packager/dialogs/inst_product_license.rb
```diff
@@ -111,10 +112,26 @@ def language_selection
...
+      # It detects whether license text is richtext or not and format it
+      # accordingly.
+      #
+      # @return [String] Formatted license text
+      def formatted_license_text
+        text = product.license(language)
+        if RICHTEXT_REGEXP =~ text
+          text
+        else
+          "<pre>#{Yast::String.EscapeTags(text)}</pre>"
```
lslezak: I'd prefer native Ruby `CGI.escapeHTML` function here, we are already [using it it elsewhere](https://github.com/yast/yast-registration/blob/c603f9634520ebe2594fd3b08bb87804974c60e6/src/lib/registration/ui/migration_selection_dialog.rb#L208).

In the long term we should drop all YaST specific code and use Ruby stdlib where possible.

imobachgs:
Thanks!

PR #264, opened 2017-08-18
fix typo in storage_include.rb

wfeldt:


src/include/packager/storage_include.rb
```diff
@@ -21,7 +21,7 @@ def ReleaseHDDUsedAsInstallationSource
         end
       end
 
-      Builins.y2milestone "install src partition #{install_src_partition}"
+      Buildins.y2milestone "install src partition #{install_src_partition}"
```
ancorgs: It's `Builtins`

PR #267, opened 2017-08-22
Multi-products licenses support

imobachgs:
This PR, which is built on top of #263, adds clients and dialogs to handle products licenses.

* `Y2Packager::Dialogs::InstProductLicense` offers a dialog to ask for license confirmation. It is used in a `Y2Packager::Clients::InstProductLicense` client.
* `Y2Packager::Clients::InstRepositoriesInitialization` initializes the packaging system. It is supposed to be used before `inst_complex_welcome` and it will set the base product if only one is available (for instance, in `openSUSE`).

### DIFF
--- a/src/lib/y2packager/clients/inst_product_license.rb
+++ b/src/lib/y2packager/clients/inst_product_license.rb
@@ -1,0 +1,30 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2017 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "yast"
+require "y2packager/dialogs/inst_product_license"
+require "y2packager/product"
+
+module Y2Packager
+  module Clients
+    class InstProductLicense
+      def main
+        dialog = Y2Packager::Dialogs::InstProductLicense.new(selected_product)
+        dialog.run
+      end
+
+      def selected_product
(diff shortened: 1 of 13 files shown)


src/lib/y2packager/clients/inst_product_license.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module Y2Packager
+  module Clients
+    # This client shows a license confirmation dialog for the base selected product
+    class InstProductLicense
+      def main
+        textdomain "installation"
+
+        if !selected_product.license?
+          return Yast::GetInstArgs.going_back ? :back : :next
```
lslezak: You can simply return `:auto` which when going back means `:back` otherwise `:next`.

src/lib/y2packager/dialogs/inst_product_license.rb
```diff
@@ -0,0 +1,147 @@
...
+        )
+      end
+
+      # Update the product's license confirmation status
+      #
+      # It will not update the status if it has not changed.
+      def update_product_confirmation
+        return if product.license_confirmed? == confirmed
+        if confirmed
+          product.confirm_license
```
lslezak: Maybe the API could be `confirm_license(confirm = true)` so you could simply call `product.confirm_license(confirmed)` instead of the condition. Or have a `confirm_license`/`deny_license`/`set_license(bool)` combo...

src/lib/y2packager/product.rb
```diff
@@ -47,6 +47,15 @@ def self.available_base_products
       Y2Packager::ProductReader.available_base_products
     end
 
+    # Returns the selected base product
+    #
+    # It assumes that at most 1 product could be selected.
+    #
+    # @return [Y2Packager::Product] Selected base product
+    def self.selected_base
+      available_base_products.find(&:selected?)
```
lslezak: I have a feeling that sooner or later we might need a separate class just for holding the base product selected by the user. With many extensions available and many derived products (like SLES4SAP, OES) and possible dependencies/conflicts the assumption of only 1 selected base product might not be valid anymore...

But that's something for later...

imobachgs: Yes, I think that this class will evolve. We will adapt the API (or split into new classes) as soon as we need to do that.

PR #272, opened 2017-08-31
Do not report errors for optional patterns (bsc#1054670)

teclator:
- [Trello card](https://trello.com/c/NwfJLH2I)
screenshot_generic-2_2017-09-01_13 24 11
screenshot_generic-2_2017-09-01_13 23 48
screenshot_leanos_2017-08-31_13 28 12

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -1051,28 +1051,11 @@ module Yast
     #-----------------------------------------------------------------------
 
     def ComputeSystemPatternList
-      pattern_list = []
-      # also add the 'laptop' selection if PCMCIA detected
-      if Arch.is_laptop || Arch.has_pcmcia
-        ["laptop", "Laptop"].each do |pat_name|
-          pat_list = Pkg.ResolvableProperties(pat_name, :pattern, "")
-          pattern_list << pat_name unless pat_list.empty?
-        end
-      end
-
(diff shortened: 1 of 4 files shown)


src/modules/Packages.rb
```diff
@@ -2490,7 +2511,7 @@ def patterns_to_install
     end
 
     def report_missing_pattern(pattern_name)
-      if optional_default_patterns.include?(pattern_name)
+      if (optional_default_patterns | resolvable_optional_patterns).include?(pattern_name)
```
lslezak: It should additionally check that the pattern is *not* listed as mandatory as well. The same pattern might be added by more components with different requirements (optional/mandatory).

If one component adds the pattern as optional and another component sets the same pattern as mandatory then the error should be still reported. With this code the pattern becomes optional globally for all components.

lslezak: It would be nice to add a test case for this...

teclator: 👍 I was currently checking exactly that, and was not sure if it should always prevail the last one at least for products (installation.xml) vs control file (control.xml). I mean, if control file defines it as default_pattern but product as optional_default pattern if it should be only as optional or in mandatory and optional lists.

After a discussion we decided the Resolvables will be just added and we should check if them are mandatory.

PR #273, opened 2017-08-31
Add widgets for product licenses

imobachgs:
Add `ProductLicense` and `ProductLicenseConfirmation` as widgets to show/confirm product licenses.

### DIFF
--- a/src/lib/y2packager/clients/inst_product_license.rb
+++ b/src/lib/y2packager/clients/inst_product_license.rb
@@ -19,11 +19,12 @@ module Y2Packager
     # This client shows a license confirmation dialog for the base selected product
     class InstProductLicense
       include Yast::I18n
+      include Yast::Logger
 
       def main
         textdomain "installation"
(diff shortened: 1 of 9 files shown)


src/lib/y2packager/clients/inst_product_license.rb
```diff
@@ -32,8 +33,36 @@ def main
...
+      # Determines whether the license should be shown
+      #
+      # @return [Boolean]
+      def show_license?
+        multi_product_media? && available_license?
+      end
+
+      # Determines whether a multi-product media is being used
+      #
+      # This client only makes sense when using a multi-product media.
```
ancorgs: Not sure about the meaning of that sentence

imobachgs: It means that this client is only used in multiproduct installation medias like LeanOS + SLES + SLED. For single product, the license will be shown in the complex welcome screen.

But we could make it easier and drop this condition. After all, the client won't be on control file of single product medias.

PR #275, opened 2017-09-05
Do not depend on system-installation() when finding base product

imobachgs:


src/lib/y2packager/product_reader.rb
```diff
@@ -41,13 +41,12 @@ def self.available_base_products
                                 installation_package: installation_mapping[prod["name"]])
       end
 
-      # only installable products
-      result.select!(&:installation_package)
+      products = find_installable(result)
```
jreidinger: why not simple
```
products = available_products

installation_mapping = installation_package_mapping
return products.first if products.size == 1 && installation_mapping.empty?

result = ...
```

imobachgs: Yes, the simpler the better.

PR #276, opened 2017-09-07
Remove copy of passwd,group,shadow from inst-sys (bsc#1057367)

teclator:
**Proposed fix for:** https://bugzilla.suse.com/show_bug.cgi?id=1057367

It seems that the YaST does not need to take care of users copy/initialization anymore before rpm installations as far systemd is taking care of it and each package should define which users and groups provides.

You can find more documentation here:

- https://en.opensuse.org/openSUSE:Packaging_guidelines#Users_and_Groups

And also the PR which discharges aaa_base's responsibility.

- https://github.com/openSUSE/aaa_base/pull/47


teclator:
It seems that the YaST does not need to take care of users copy/initialization anymore before rpm installations as far systemd is taking care of it and each package should define which users and groups provides.

You can find more documentation here:

- https://en.opensuse.org/openSUSE:Packaging_guidelines#Users_and_Groups

And also the PR which discharges aaa_base's responsibility.

- https://github.com/openSUSE/aaa_base/pull/47

@thkukuk could you take a look, it would be nice to have your comments about it.

thkukuk:
Fine with me.

teclator:
@lslezak  @thkukuk thnx for review ;)

PR #277, opened 2017-09-12
Replace info.txt beta file by README.BETA

teclator:
- [Trello Card](https://trello.com/c/EQ5N9yjI/1064-5-sle-15-show-beta-warning-if-readmebeta-is-found-for-products-modules-and-extensions)

### DIFF
--- a/src/modules/InstShowInfo.rb
+++ b/src/modules/InstShowInfo.rb
@@ -4,7 +4,7 @@ require "yast"
 
 # Yast namespace
 module Yast
-  # Show /info.txt (if present) in a popup
+  # Show a given info file (if present) in a popup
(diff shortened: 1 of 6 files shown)


src/modules/InstShowInfo.rb
```diff
@@ -17,7 +17,7 @@ def main
       Yast.import "Label"
     end
 
-    # @param [String] info_file (/info.txt" - Copied from inst media to inst-sys by linuxrc)
+    # @param [String] info_file path to be shown
```
jreidinger: is it really info file or beta one? Maybe not worth it to change API name

teclator: yep, by now I decided to not change the API name

PR #278, opened 2017-09-13
show in url also product dir if it is defined

jreidinger:
snimek obrazovky_2017-09-13_11-00-46

zypper lr output:
```
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias                    | Name                               | Enabled | GPG Check | Refresh
--+--------------------------+------------------------------------+---------+-----------+--------
1 | SUSE_SLE-15_GA           | Code 15 SP0 GA platform (standard) | Yes     | (r ) Yes  | No     
2 | leanos15-15-0            | leanos15-15-0                      | Yes     | (r ) Yes  | No     
3 | multi_modules_Basesystem | multi modules (/Basesystem)        | Yes     | (r ) Yes  | Yes    
4 | multi_modules_Legacy     | multi modules (/Legacy)            | Yes     | (r ) Yes  | Yes    
```

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Sep 14 14:30:50 UTC 2017 - jreidinger@suse.com
+
+- show product path in url for multi product medias
+- for multi product medias add product path to repository name
+  (fate#323450)
+- 4.0.1
+
+-------------------------------------------------------------------
 Tue Sep 12 10:19:41 UTC 2017 - knut.anderssen@suse.com
 
 - Replace /info.txt by /README.BETA as beta file to be shown
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,8 +1,8 @@
 -------------------------------------------------------------------
 Thu Sep 14 14:30:50 UTC 2017 - jreidinger@suse.com
 
-- show product path in url for multi product medias
-- for multi product medias add product path to repository name
+- show product path in url for multi product media
+- for multi product media add product path to repository name
   (fate#323450)
 - 4.0.1
 
--- a/src/include/packager/repositories_include.rb
+++ b/src/include/packager/repositories_include.rb
@@ -210,7 +210,7 @@ module Yast
           end
           alias_name = ""
           if force_alias == ""
-            # replace " " -> "_" (avoid spaces in .repo file name) and remove shell unfriendly '/()'
+            # replace " " -> "_" (avoid spaces in .repo file name) and remove shell unfriendly chars
             alias_name = name.tr(" ", "_").delete(SHELL_UNFRIENDLY)
             alias_orig = alias_name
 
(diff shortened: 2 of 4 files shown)


lslezak:
Looks OK, thanks!

lslezak:
And what is the alias in this case? What does `zypper lr` display?

The alias is created from the name later (https://github.com/yast/yast-packager/pull/278/files#diff-3181dd45a1dd0f7bbc892e69925b0989R211), I'm not sure how the slash (`/`) in the alias is handled. Alias is used for the `*.repo` file name is it can be a bit tricky...

jreidinger:
@lslezak this is output

```
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias                       | Name                               | Enabled | GPG Check | Refresh
--+-----------------------------+------------------------------------+---------+-----------+--------
1 | SUSE_SLE-15_GA              | Code 15 SP0 GA platform (standard) | Yes     | (r ) Yes  | No     
2 | leanos15-15-0               | leanos15-15-0                      | Yes     | (r ) Yes  | No     
3 | multi_product_(/Basesystem) | multi product (/Basesystem)        | Yes     | (r ) Yes  | Yes    
4 | multi_product_(/Legacy)     | multi product (/Legacy)            | Yes     | (r ) Yes  | Yes 
```

maybe we can strip there at least `()`

lslezak:
I'd remove all `()/`, using these on command line requires escaping, that's ugly.

jreidinger:
@lslezak in general or only in this case? ( multi product repo )

lslezak:
In general, those characters should not be used for any repository. So remove it in all cases.

src/include/packager/repositories_include.rb
```diff
@@ -204,8 +207,8 @@ def createSourceImpl(url, plaindir, download, preffered_name, force_alias)
           end
           alias_name = ""
           if force_alias == ""
-            # replace " " -> "_" (avoid spaces in .repo file name)
-            alias_name = Builtins.mergestring(Builtins.splitstring(name, " "), "_")
+            # replace " " -> "_" (avoid spaces in .repo file name), and remove shell unfriendly '/()'
```
lslezak: NP: Maybe we could remove more characters like `*&<>'"`...

src/include/packager/repositories_include.rb
```diff
@@ -204,8 +210,8 @@ def createSourceImpl(url, plaindir, download, preffered_name, force_alias)
           end
           alias_name = ""
           if force_alias == ""
-            # replace " " -> "_" (avoid spaces in .repo file name)
-            alias_name = Builtins.mergestring(Builtins.splitstring(name, " "), "_")
+            # replace " " -> "_" (avoid spaces in .repo file name) and remove shell unfriendly '/()'
```
lslezak: NP: Update the comment as well, `remove shell unfriendly characters` sounds good enough.

package/yast2-packager.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Sep 14 14:30:50 UTC 2017 - jreidinger@suse.com
+
+- show product path in url for multi product medias
```
lslezak: Bad English: singular is `medium`, plural is `media`
(Fix the line below as well.)

PR #280, opened 2017-09-14
Show the beta file even if no license is present (bsc#1047060)

teclator:
- [Trello Card](https://trello.com/c/EQ5N9yjI/1064-5-sle-15-show-beta-warning-if-readmebeta-is-found-for-products-modules-and-extensions)

It just shows the BETA file in case of present, but we probably will change the way a repository is added so, by now, I just have adapted the old way.

show_beta

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -11,7 +11,7 @@ module Yast
     include Yast::Logger
 
     DOWNLOAD_URL_SCHEMA = ["http", "https", "ftp"].freeze
-    INFO_FILE = "/README.BETA".freeze
+    README_BETA = "/README.BETA".freeze
 
     def main
       Yast.import "Pkg"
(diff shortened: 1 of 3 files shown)


src/modules/ProductLicense.rb
```diff
@@ -1182,11 +1182,10 @@ def SearchForLicense_NormalRunBaseProduct(_src_id, fallback_dir)
     def SearchForLicense_AddOnProduct(src_id, _fallback_dir)
       Builtins.y2milestone("Getting license info from repository %1", src_id)
 
-      @info_file = Pkg.SourceProvideDigestedFile(
+      @info_file = Pkg.SourceProvideOptionalFile(
         src_id, # optional
         1,
-        "/media.1" + INFO_FILE,
-        true
+        INFO_FILE
```
lslezak: Actually I find it a bit confusing, in the registration we remove the info file handling completely and here the code is left, just changing the `INFO_FILE` value to `README.BETA`. 

That's confusing, the methods/variables mean something completely different than the name suggests. I'd rename the variables/methods (except the `InstShowInfo` module methods not not break the API)...

teclator: It makes sense.

PR #282, opened 2017-09-15
Do not show license agreement twice

imobachgs:
This should fix [bsc#1058503](https://bugzilla.suse.com/show_bug.cgi?id=1058503).


imobachgs:
Thanks!

PR #283, opened 2017-09-18
Add on product selection for multi-module media

lslezak:
- Display a selection dialog to choose the products to install when using a multi-module medium
- See https://trello.com/c/Hg8HZTYm for details (e.g. a testing repository)
- Note: Maybe reviewing per commits is better as there is some cleanup and refactoring which makes reading the change more difficult...
- Note: The coverage increase of +0.8% does not sound that much but actually the changed `repositories_include.rb` file improved from 17.1% to 60.5%

[TODO: add screenshot]

### DIFF
--- a/src/include/packager/repositories_include.rb
+++ b/src/include/packager/repositories_include.rb
@@ -201,7 +201,7 @@ module Yast
           URL.HidePassword(url)
         )
 
-        if Builtins.regexpmatch(url, "\\.iso$")
+        if url.end_with?(".iso")
           parsed_url = URL.Parse(url)
           scheme2 = Builtins.tolower(Ops.get_string(parsed_url, "scheme", ""))
 
@@ -223,10 +223,7 @@ module Yast
           _("Change the URL and try again?")
         )
 
-        tryagain = Popup.YesNo(msg)
-        return :again if tryagain
-
-        return :cancel
+        Popup.YesNo(msg) ? :again : :cancel
       else
         Progress.NextStage
         license_accepted = true
@@ -253,7 +250,7 @@ module Yast
         # relese (unmount) the medium
         Pkg.SourceReleaseAll
 
-        return license_accepted ? :ok : :abort
+        license_accepted ? :ok : :abort
       end
     end
 
--- a/src/include/packager/repositories_include.rb
+++ b/src/include/packager/repositories_include.rb
@@ -8,7 +8,7 @@ module Yast
     include Yast::Logger
 
     # constant Plaindir
-    PLAINDIR_TYPE = "Plaindir"
+    PLAINDIR_TYPE = "Plaindir".freeze
 
     def initialize_packager_repositories_include(_include_target)
       Yast.import "Pkg"
@@ -35,13 +35,13 @@ module Yast
     end
 
     def LicenseAccepted(id)
-      ret = AddOnProduct.AcceptedLicenseAndInfoFile(id)
-      ret
+      AddOnProduct.AcceptedLicenseAndInfoFile(id)
     end
 
     # shell unfriendly characters we want to remove from alias, so it is easier to use with zypper
     SHELL_UNFRIENDLY = "()/!'\"*?;&|<>{}$#`".freeze
 
+    # Create a new repository or service.
     # @param url [String] repository or service URL
     # @param plaindir [Boolean] true to use "PlainDir" format (no repository
     #   metadata present at the URL)
@@ -50,9 +50,14 @@ module Yast
     # @param force_alias [String] alias for the new repository, if a repository
     #   with this alias already exists then it is overwritten, use empty string ""
     #   to generate an unique alias
+    # @return [Symbol] the result
+    #   :ok => successfully added
+    #   :again => failed, but user wants to edit the URL and try again
+    #   :cancel => failed, don't retry
+    #   :abort => repository added successfully, but user rejected the license
     def createSourceImpl(url, plaindir, download, preffered_name, force_alias)
       log.info("createSource: #{URL.HidePassword(url)}, plaindir: #{plaindir}," \
-        "download: #{download}, name: #{preffered_name}")
+        "download: #{download}, name: #{preffered_name}, force_alias: #{force_alias}")
 
       if url.nil? || url.empty?
         Builtins.y2error(-1, "Empty URL! Backtrace:")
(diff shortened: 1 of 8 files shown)


src/include/packager/repositories_include.rb
```diff
@@ -25,329 +34,179 @@ def initialize_packager_repositories_include(_include_target)
...
+      Builtins.y2milestone("Probed service type: %1", service_type)
 
-          if Ops.greater_than(Builtins.size(p_elems), 1)
-            url_path = Ops.get(
-              p_elems,
-              Ops.subtract(Builtins.size(p_elems), 1),
-              fallback
-            )
+      # create a new service if a service is detected at the URL
+      if !service_type.nil? && service_type != "NONE"
```
jreidinger: `if ![nil, "NONE"].include?(service_type)`

src/include/packager/repositories_include.rb
```diff
@@ -25,329 +34,179 @@ def initialize_packager_repositories_include(_include_target)
...
+        # nothing selected or pressed abort/cancel/close/back/...
+        return :cancel if found_products.empty? || ui != :next
+      end
 
-              repo_type = @plaindir_type
-              Builtins.y2warning(
-                "Probing has failed, using Plaindir repository type."
-              )
+      found_products.each do |product|
+        next if enter_again
```
jreidinger: this smells. What is logic there and why break caant be used at line 119?

src/include/packager/repositories_include.rb
```diff
@@ -25,329 +34,179 @@ def initialize_packager_repositories_include(_include_target)
...
+        name = !preffered_name.nil? && preffered_name != "" ? preffered_name : product.name
```
jreidinger: if you know it is not nil, then `empty?`shoyld be preferred

src/include/packager/repositories_include.rb
```diff
@@ -25,329 +34,179 @@ def initialize_packager_repositories_include(_include_target)
...
-          # popup message part 2
-          msg = Ops.add(
-            Ops.add(msg, "\n\n"),
-            _("Change the URL and try again?")
-          )
+      # broken repository or wrong URL - enter the URL again
+      if enter_again
+        Pkg.SourceReleaseAll
+        return :again
+      end
```
jreidinger: btw why not do this when setting enter again to true?

src/include/packager/repositories_include.rb
```diff
@@ -25,329 +34,179 @@ def initialize_packager_repositories_include(_include_target)
...
+              "do_refresh"  => download
+            }
+            @sourceStatesOut << sourceState
           end
+        end
 
-          # relese (unmount) the medium
-          Pkg.SourceReleaseAll
+        # relese (unmount) the medium
+        Pkg.SourceReleaseAll
```
jreidinger: well, I expect that this can be in `ensure` and in such case work-flow in this method can be simplified by `return :abort` on line 184

src/include/packager/repositories_include.rb
```diff
@@ -469,5 +327,185 @@ def autorefresh_for?(url)
...
       autorefresh
     end
+
+  private
+
+    # scan the repository URL and return the available products
+    # @return [Array<Y2Packager::ProductLocation>] Found products
+    def scan_products(expanded_url, original_url)
+      new_repos = Pkg.RepositoryScan(expanded_url)
+      found_products = new_repos.map { |r| Y2Packager::ProductLocation.new(r[0], r[1]) }
```
jreidinger: `found_products = new_repos.map { |(name, dir)| Y2Packager::ProductLocation.new(name, dir) }` is more readable.

src/include/packager/repositories_include.rb
```diff
@@ -469,5 +327,185 @@ def autorefresh_for?(url)
...
+              fallback
+            )
+
+            url_path = fallback if url_path.nil? || url_path == ""
+          end
+        elsif url_path == "/"
+          url_path = fallback
+        end
+
+        found_products << Y2Packager::ProductLocation.new(url_path, "/")
```
jreidinger: own method? this is already quite long. so whole this if can be own method and it can look like
```
found_products = [single_product(original_url)] if found_products.empty?

log.info("Found products: #{found_products}")

return found_products
```

src/include/packager/repositories_include.rb
```diff
@@ -469,5 +327,185 @@ def autorefresh_for?(url)
...
+      aliases = @serviceStatesOut.map { |s| s["alias"] }
+
+      # service alias must be unique
+      # if it already exists add "_<number>" suffix to it
+      idx = 1
+      alias_name = "service"
+      while aliases.include?(alias_name)
+        alias_name = "service_#{idx}"
+        idx += 1
+      end
```
jreidinger: DRY warning , already seen at line 388. Can it be own method?

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -0,0 +1,115 @@
...
+            ))
+        )
+      end
+
+      def read_user_selection
+        selected_items = Yast::UI.QueryWidget(Id("addon_repos"), :SelectedItems)
+
+        self.selected_products = selected_items.map do |s|
+          products.find { |p| p.name == s }
+        end
```
jreidinger: what about 
```
self.selected_products = product.select { |p| selected_items.include?(p.name) }
```

src/lib/y2packager/product_location.rb
```diff
@@ -0,0 +1,32 @@
...
+    attr_reader :name
+    # @return [String] User selected products
+    attr_reader :dir
+
+    # Constructor
+    #
+    # @param name [String] Product name
+    # @param dir [String] Location (path starting at the media root)
+    def initialize(name, dir)
+      super()
```
jreidinger: hmm, what?

src/lib/y2packager/product_location.rb
```diff
@@ -0,0 +1,32 @@
...
+      @name = name
+      @dir = dir
+    end
```
jreidinger: this looks like exactly `Struct.new(:name, :dir)`

test/repositories_include_test.rb
```diff
@@ -2,12 +2,25 @@
...
-class RepositoryIncludeTester
-  extend Yast::I18n
-  Yast.include self, "packager/repositories_include.rb"
+module Yast
+  class RepositoryIncludeTesterClass < Module
+    extend Yast::I18n
+  end
+
+  def main
+    Yast.include self, "packager/repositories_include.rb"
```
jreidinger: so you include to whole Yast namespace this repositories include? Why not include it to RepositoryIncludeTesterClass?

src/include/packager/repositories_include.rb
lslezak: Um, this would need more analysis how the code is called from outside, although it looks strange I do not date to touch this now...

src/include/packager/repositories_include.rb
lslezak: Again, I'm not sure about the usage from outside, I'd just keep it as it is.

PR #287, opened 2017-09-28
[Review] Request from 'schubi2' @ 'yast/yast-packager/review_170928_disable_remote_installation'

schubi2:
  * 957f2c35 disabling vnc, shh,.... if it is not supported

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -975,16 +975,16 @@ module Yast
       missing_remote_packages.flatten!
       unless missing_remote_packages.empty?
         error_string = format(
-          _("Cannot support %s remote access in the installed system due missing packages \n%s. \n" \
-            "It will be disabled."),
+          _("Cannot support %s remote access in the installed system due missing " \
+            "packages \n%s. \nIt will be disabled."),
           @missing_remote_kind.join(", "), @missing_remote_packages.join(", "))
         if Mode.auto
           error_string << " \n"
           error_string << _("But the AutoYaST installation will be still finished automatically " \
             "without any user interaction.")
         end
-        log.warn ("Cannot support #{@missing_remote_kind.join(", ")} remote access in the " \
-          "installed system due missing packages #{@missing_remote_packages.join(", ")}")
+        log.warn("Cannot support #{@missing_remote_kind} remote access in the " \
+          "installed system due missing packages #{@missing_remote_packages}")
         return error_string
       end
       ""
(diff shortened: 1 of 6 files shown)


src/modules/Packages.rb
```diff
@@ -2688,6 +2759,24 @@ def create_baseproduct_symlink
...
+    # Checking if a package will be installed or is already installed
+    # on a system and will not be deleted.
+    # @param [String] package name
+    # @return [Boolean] true if the package will be on the installed system
+    def pkg_will_be_installed(tag)
+      provides = Pkg.PkgQueryProvides(tag)
+      # e.g.: [["kernel-bigsmp", :CAND, :NONE], ["kernel-default", :CAND, :CAND],
+      # ["kernel-default", `BOTH, :INST]]
+      log.info("provides: #{provides}")
+      if provides.any? { |p| p[2] != :NONE }
```
lslezak: I'd avoid hardcoding the result values and use directly the `any?` result:
```ruby
ret = provides.any? { |p| p[2] != :NONE }
log.info("#{tag} will #{ret ? "" : "not "}be installed")
return ret
```

test/packages_test.rb
```diff
@@ -914,6 +914,168 @@ def product(properties = {})
...
+      allow(Yast::Packages).to receive(:ssh_packages).and_return(ssh_packages)
+      allow(Yast::Packages).to receive(:braille_packages).and_return(braille_packages)
+    end
+
+    let(:vnc_packages) { %w(some-vnc-packages) }
+    let(:remote_x11_packages) { %w(some-x11-packages) }
+    let(:ssh_packages) { %w(openssh iproute2) }
+    let(:braille_packages) { %w(sbl) }
+
+    context "on a boring local regular installation" do
```
lslezak: I do a lot of boring installations during testing... :stuck_out_tongue_closed_eyes:

src/modules/Packages.rb
```diff
@@ -927,23 +936,66 @@ def architecturePackages
...
+          @missing_remote_kind << "braille"
+        end
+      end
       if Linuxrc.usessh
-        # "ip" tool is needed by the YaST2.ssh start script (bnc#920175)
-        tags.concat(["openssh", "iproute2"])
+        missing = ssh_packages.reject { |tag| pkg_will_be_installed(tag) }
+        unless missing.empty?
+          @missing_remote_packages << missing
+          @missing_remote_kind << "ssh"
```
lslezak: As this is also displayed in the UI, using the upper case variant is better: `SSH`, `VNC`,...

PR #291, opened 2017-10-16
Default desktop cleaning

jreidinger:


src/modules/DefaultDesktop.rb
```diff
@@ -25,153 +26,47 @@ def main
...
-      case key
-      when "order"
-        99
-      when "desktop"
-        "unknown"
-      else
-        ""
-      end
-    end
+    DEFAULT_VALUES.default_proc = Proc.new { "" }
```
lslezak: NP: The `default_proc` is not common in the YaST code, I'd suggest add a comment like
```Ruby
# define the default value for missing keys
```

src/modules/DefaultDesktop.rb
```diff
@@ -25,153 +26,47 @@ def main
...
 
-        if any_desktops_from_cf != "" && any_desktops_from_cf != ""
-          desktops_from_cf = Convert.convert(
-            any_desktops_from_cf,
-            from: "any",
-            to:   "list <map>"
-          )
-        end
+        log.info "desktops from control file #{desktops_from_cf.inspect}"
+        desktops_from_cf = [] unless desktops_from_cf.is_a?(::Array)
```
lslezak: NP: I'd at least log a warning/error that the control file contains an invalid value which is ignored.

src/modules/DefaultDesktop.rb
```diff
@@ -181,12 +76,36 @@ def Init
...
+        "label_id" => desktop_label,
+        "label"    => ProductControl.GetTranslatedText(desktop_label)
+      }
+      ["desktop", "logon", "cursor", "order"].each do |key|
+        one_desktop[key] = control_file_entry[key] || DEFAULT_VALUES[key]
+      end
+      ["packages", "patterns"].each do |key|
+        one_desktop[key] = (control_file_entry[key] || "").split
+      end
+      # 'icon' in optional
```
lslezak: Err, I guess it should be `'description' is optional`...

lslezak:
Plus Rubocop failed...

src/modules/DefaultDesktop.rb
jreidinger: well, main use case is nil when it is not defined. Other cases can be already seen in previous log.info and I consider it quite minor.

PR #295, opened 2017-10-27
Fixed installing additional products at upgrade (bsc#1065485)

lslezak:
# Product Selection Fix

This fixes yet another instance of the *all available products* bug, see [bsc#1065485](https://bugzilla.suse.com/show_bug.cgi?id=1065485).

### The Original Issue

Originally at SLES12-SP3 upgrade YaST additionally installed also the other products like SLED which causes dependency issues (the products conflicts):

media_upgrade_broken_products

### With the Fix

With this fix no additional products are installed. The products should be upgraded using the *-release RPM packages on the new media, we do not need to select anything to install.

media_upgrade_correct_products

# Testing Client

I have added my small testing client used for developing the module selection dialog. I think it can be useful for somebody else later so let's share it.

### DIFF
--- a/devel/README.md
+++ b/devel/README.md
@@ -1,4 +1,4 @@
 # Development Code
 
-This directory contains code examples which is can be useful for development.
+This directory contains code examples which can be useful for development.
 They are not intented for regular use by end users.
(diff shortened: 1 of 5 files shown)


devel/README.md
```diff
@@ -0,0 +1,4 @@
+# Development Code
+
+This directory contains code examples which is can be useful for development.
```
imobachgs: *which ~is~ can be useful*

PR #296, opened 2017-11-08
Select available product on single product media

imobachgs:
It should fix these bugs:

* [bsc#1066709](https://bugzilla.suse.com/show_bug.cgi?id=1066709)
* [bsc#1065172](https://bugzilla.suse.com/show_bug.cgi?id=1065172)

Manually tested with the Storage-NG ISO (https://yast-openqa.suse.cz/tests/5331).

### DIFF
--- a/src/lib/y2packager/clients/inst_repositories_initialization.rb
+++ b/src/lib/y2packager/clients/inst_repositories_initialization.rb
@@ -73,13 +73,22 @@ module Y2Packager
 
       # Adjust product selection
       #
-      # All products are selected by default. So if there is more than 1, we should unselect
-      # them all. The user will select one later.
+      # During installation, all products are selected by default. So if there
+      # is more than 1, we should unselect them all. The user will select one
+      # later.
       #
-      # See https://github.com/yast/yast-packager/blob/7e1a0bbb90823b03c15d92f408036a560dca8aa3/src/modules/Packages.rb#L1876
+      # On the other hand, if only one product is available, we should make
+      # sure that it is selected for installation because, during upgrade, it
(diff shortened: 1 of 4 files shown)


teclator:
For me looks good just wait for @lslezak second opinion.

src/lib/y2packager/clients/inst_repositories_initialization.rb
```diff
@@ -73,13 +73,22 @@ def init_installation_repositories
...
+      # sure that it is selected for installation because, during upgrade, it
+      # is not automatically selected.
+      #
+      # @see https://github.com/yast/yast-packager/blob/7e1a0bbb90823b03c15d92f408036a560dca8aa3/src/modules/Packages.rb#L1876
+      # @see https://github.com/yast/yast-packager/blob/fbc396df910e297915f9f785fc460e72e30d1948/src/modules/Packages.rb#L1905
       def adjust_base_product_selection
-        return unless products.size > 1
-        products.each(&:restore)
+        if products.size == 1
+          products.first.select
```
lslezak: Um, could we skip the product selection at upgrade? It's quite strange to select it if there is only one available. If there are multiple products the correct product will be upgraded indirectly via the `*-release` package upgrade.

I think we should do the same with just one product so it behaves the same way. The less difference the better.

lslezak: According to @imobachgs it's needed because of the license later.

imobachgs: Just for the records, it is needed in order to show the license (discussed on IRC).

PR #286, opened 2017-09-25
Remove xinetd from the list of VNC packages.

michalsrb:
The VNC service switched from xinetd to systemd.

This comes after:
https://github.com/yast/yast-network/pull/565

PR #297, opened 2017-11-09
Remove xinetd from the list of VNC packages.

teclator:
- It will part of this incoming [Trello Card](https://trello.com/c/yTCpqUtX/1728-sles15-p3-1058820-yast2-remote-administration-vnc-complains-about-missing-servicevnc-httpd)

- Fixed conflicts and added changelog to #286 

- This is needed by https://github.com/yast/yast-network/pull/565 (I have been working on some cleanup, so hope to create a PR which supersedes also original one)

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,3 +1,8 @@
+Thu Nov  9 14:45:15 UTC 2017 - knut.anderssen@suse.com
+
+- VNC does not depend on xinetd anymore. (bsc#1067169)
+- 4.0.18
+
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


package/yast2-packager.changes
```diff
@@ -1,3 +1,8 @@
+Thu Nov  9 14:45:15 UTC 2017 - knut.anderssen@suse.com
```
lslezak: IIRC the change entry should start with the `-----` separator on the first line.

teclator: yep, I realized it yesterday when cheking, an also the bug reference is wrong needs to fix it, but was working in yast2-network (remote) changes related to this PR.

Thnx for checking!!

PR #299, opened 2017-12-01
[Review] Request from 'schubi2' @ 'yast/yast-packager/review_171201_adding_update_repo_ay'

schubi2:
  * d2564872 packaging
  * 4353be6a allow AY to add a product

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -847,13 +847,7 @@ module Yast
           "Found list of add-on products to preselect: %1",
           @add_on_products_list
         )
-        if Mode.auto
-          Builtins.y2warning("This is an AutoYaST installation. "\
-            "Only Add-on products will be added which have been defined "\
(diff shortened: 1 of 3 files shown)


src/modules/Packages.rb
```diff
@@ -847,13 +847,7 @@ def InitializeAddOnProducts
...
           @add_on_products_list
         )
-        if Mode.auto
-          Builtins.y2warning("This is an AutoYaST installation. "\
-            "Only Add-on products will be added which have been defined "\
-            "in the add-on section of the AutoYaST configuration file.")
-        else
-          AddOnProduct.AddPreselectedAddOnProducts(@add_on_products_list)
-        end
+        AddOnProduct.AddPreselectedAddOnProducts(@add_on_products_list)
```
lslezak: This will also add the products from `/add_on_products.xml` or (plain text `/add_on_products`) file.

This should be documented and possibly mentioned in the release notes (an incompatible change).

PR #300, opened 2017-12-07
Fix disable local repos

imobachgs:
Now we only consider installed products when deciding whether to disable a repository in order to work with multi-product media.

[bsc#1071702](https://bugzilla.suse.com/show_bug.cgi?id=1071702)

### DIFF
--- a/src/lib/packager/clients/pkg_finish.rb
+++ b/src/lib/packager/clients/pkg_finish.rb
@@ -145,9 +145,9 @@ module Yast
     #
     # @return [Array<Y2Packager::Repository>] List of disabled repositories
     def disable_local_repos
-      candidates_repos, remote_repos = *::Y2Packager::Repository.enabled.partition(&:local?)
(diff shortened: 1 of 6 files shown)


src/lib/packager/clients/pkg_finish.rb
```diff
@@ -145,14 +145,14 @@ def backup_all_target_sources
     #
     # @return [Array<Y2Packager::Repository>] List of disabled repositories
     def disable_local_repos
-      candidates_repos, other_repos = *::Y2Packager::Repository.enabled.partition(&:local?)
-      products = other_repos.map(&:products).flatten.uniq
+      candidates_repos, remote_repos = *::Y2Packager::Repository.enabled.partition(&:local?)
```
lslezak: NP: I think you could rename `candidates_repos` to `local_repos` so we have `local_repos` and `remote_repos` pair.

PR #301, opened 2017-12-11
[Review] Request from 'schubi2' @ 'yast/yast-packager/review_171211_bug_1061754'

schubi2:
  * 9e7d6f1d packaging
  * 5956f429 AY: warn user if second stage is not possible

### DIFF
--- a/src/lib/packager/clients/software_proposal.rb
+++ b/src/lib/packager/clients/software_proposal.rb
@@ -30,7 +30,12 @@ module Yast
 
       Yast.import "Packages"
       Yast.import "Language"
+      Yast.import "Mode"
+      # The following imports are modules which are used in the inst-sys
+      # only. So they have not to be required by the package because
+      # the software proposal will be called in inst-sys only.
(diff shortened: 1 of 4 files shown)


src/lib/packager/clients/software_proposal.rb
```diff
@@ -93,6 +95,15 @@ def make_proposal(flags)
         end
       end
 
+      if Yast::Mode.auto
+        # AY: Checking if second stage is needed and the environment has been setup.
+        error_message = Yast::AutoinstData.autoyast_second_stage_error
+        unless error_message.empty?
+          @ret["warning"] ? @ret["warning"] << "\n" : @ret["warning"] = ""
```
mvidner: This is an abuse of the ternary expression operator to perform a conditional statement. It happens to do the right thing but I had to consult [an operator precedence table](https://stackoverflow.com/questions/21060234/ruby-operator-precedence-table) and run `ruby-parse` to be sure. Please use `if`, `else`, `end`.

PR #302, opened 2017-12-11
Fix product status

imobachgs:
* Fixes [bsc#1072063](https://bugzilla.suse.com/show_bug.cgi?id=1072063)
* The bug was caused because we have several `SLES` instance products (one as `:removed` and another one as `:selected`).
* Trying to cache product status can be dangerous, so I've modified product class to expose a `status?` method.
* I've fixed documentation and added `rake check:doc` to Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,4 @@ script:
   # the "yast-travis-ruby" script is included in the base yastdevel/ruby image
   # see https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby
   - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-packager-image yast-travis-ruby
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -12,3 +12,4 @@ script:
   # the "yast-travis-ruby" script is included in the base yastdevel/ruby image
   # see https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby
   - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-packager-image yast-travis-ruby
+  - docker run -it -e TRAVIS=1 -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" yast-packager-image rake check:doc
```
lslezak: Thanks for the improvement! 👍

PR #306, opened 2018-01-08
Optionally skip package installation for addons (related to bsc#1074766)

lslezak:
- Display the installation summary tab by default in the package
  manager when adding addons

- 4.0.25

### DIFF
--- a/src/modules/AddOnProduct.rb
+++ b/src/modules/AddOnProduct.rb
@@ -567,8 +567,8 @@ module Yast
       )
 
       # display pattern the dialog when there is a pattern provided by the addon
-      # otherwise use search mode
-      mode = AnyPatternInRepo() ? :patternSelector : :searchMode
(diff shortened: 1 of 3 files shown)


src/modules/AddOnProduct.rb
```diff
@@ -818,8 +818,10 @@ def RegisterAddOnProduct(src_id)
     # Do installation of the add-on product within an installed system
     # srcid is got via AddOnProduct::src_id
     #
+    # @param install_packages [Boolean] install the selected packages,
+    #   can be optionally disabled
     # @return [Symbol] the result symbol from wizard sequencer
-    def DoInstall
+    def DoInstall(install_packages = true)
```
jreidinger: what about using named argument? In general I do not like much boolean params. Also documentation is a bit lacking as this happen only when it missing installation.xml, not?

PR #308, opened 2018-01-08
Ignore the partition with the installer (bsc#1073696)

lslezak:
- Ignore the partitions mounted before the installer is started
- E.g. ignore the HDD installation source or the user mounts for remote logging
- See https://bugzilla.suse.com/show_bug.cgi?id=1073696
- Tested manually with manually mounted `/var/log` before running the installer - it was correctly ignored
- Added a simple unit test
- 4.0.27

### DIFF
--- a/src/modules/SpaceCalculation.rb
+++ b/src/modules/SpaceCalculation.rb
@@ -1052,7 +1052,23 @@ module Yast
     # @return [Array<Storage::Filesystem>]
     def target_filesystems
       filesystems = Y2Storage::Filesystems::BlkFilesystem.all(staging_devicegraph)
-      filesystems.select! { |fs| fs.mountpoint && fs.mountpoint.start_with?("/") }
+      filesystems.select! do |fs|
+        # Ignore the devices mounted before starting the installer (e.g. the
+        # installation repository mounted by linuxrc when installing from HDD or
+        # the user mounted devices like for remote logging). Such devices will
+        # not be saved in the final /etc/fstab therefore check that flag.
+        # Check this only in the initial installation (as the non-fstab values
+        # will be missing in "/mnt"), in installed system they will stay available
+        # at "/".
+        # TODO: use a better API when provided by the libstorage-ng wrapper
+        if fs.mountpoint
+          log.debug("#{fs.mountpoint.inspect} in fstab: " \
+            "#{fs.to_storage_value.mount_point.in_etc_fstab?}")
+        end
+
+        fs.mountpoint && fs.mountpoint.start_with?("/") &&
+          (!Stage.initial || fs.to_storage_value.mount_point.in_etc_fstab?)
+      end
       filesystems.reject! { |fs| TARGET_FS_TYPES_TO_IGNORE.include?(fs.type) }
       filesystems
     end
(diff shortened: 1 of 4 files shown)


src/modules/SpaceCalculation.rb
```diff
@@ -1052,7 +1052,10 @@ def size_from_string(size_str)
     # @return [Array<Storage::Filesystem>]
     def target_filesystems
       filesystems = Y2Storage::Filesystems::BlkFilesystem.all(staging_devicegraph)
-      filesystems.select! { |fs| fs.mountpoint && fs.mountpoint.start_with?("/") }
+      filesystems.select! do |fs|
+        # ignore the installation repository mounted by linuxrc (when installing from HDD)
```
jreidinger: please add bnc here

lslezak:
BTW another use case is [remote logging](https://en.opensuse.org/SDB:YaST_remote_logging_during_installation) during installation.

lslezak:
IMHO the correct fix would be to clean the staging tree after initializing libstorage-ng to ignore everything what was mounted before YaST started. But we have to do it probably only in the inst-sys (`Stage.initial`) and be careful at upgrade...

lslezak:
Just for the record: The suggestion from Arvin and Ancor is to use
```ruby
fs.mountpoint && fs.mountpoint.start_with?("/") && fs.to_storage_value.mountpoints[0].in_etc_fstab
```
(Not tested!)

ancorgs:
Just to clarify, it looks so weird because https://trello.com/c/NYy7AklQ/577-use-new-libstorage-ng-api-for-mount-points is still not implemented.

lslezak:
Update: `fs.to_storage_value.mountpoints[0].in_etc_fstab` fails with `undefined method 'in_etc_fstab' for "/var/log":String` error. It seems that the `mountpoints` Array contains Strings, not some storage objects...

src/modules/SpaceCalculation.rb
```diff
@@ -1052,7 +1052,23 @@ def size_from_string(size_str)
     # @return [Array<Storage::Filesystem>]
     def target_filesystems
       filesystems = Y2Storage::Filesystems::BlkFilesystem.all(staging_devicegraph)
-      filesystems.select! { |fs| fs.mountpoint && fs.mountpoint.start_with?("/") }
+      filesystems.select! do |fs|
+        # Ignore the devices mounted before starting the installer (e.g. the
```
ancorgs: I miss the bug number somewhere in this comment.

src/modules/SpaceCalculation.rb
```diff
@@ -1052,7 +1052,23 @@ def size_from_string(size_str)
...
+        # installation repository mounted by linuxrc when installing from HDD or
+        # the user mounted devices like for remote logging). Such devices will
+        # not be saved in the final /etc/fstab therefore check that flag.
+        # Check this only in the initial installation (as the non-fstab values
+        # will be missing in "/mnt"), in installed system they will stay available
+        # at "/".
+        # TODO: use a better API when provided by the libstorage-ng wrapper
```
ancorgs: As discussed in irc? Why not create that better API right away? What about adding, e.g., a `Mountable#persistent_mount_point?` method to yast-storage-ng? Of course, I'm sure you can come up with a better name. :wink:

ancorgs: yast2-storage-ng is there to evolve and adapt to the needs of the other YaST modules, not to be worked around. :-)

PR #309, opened 2018-01-11
Lower the priority of the initial installation repository (bsc#1071742)

lslezak:
- To prefer the packages from the other media.
- See https://bugzilla.suse.com/show_bug.cgi?id=1071742
- 4.0.27

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -1741,6 +1741,10 @@ module Yast
         end
       end
 
+      # Lower the priority of the initial installation repository to prefer
+      # the packages from the other media (bsc#1071742)
+      lower_repo_priority(initial_repository)
+
       # BNC #481828: Using LABEL from product
       AdjustSourcePropertiesAccordingToProduct(@base_source_id)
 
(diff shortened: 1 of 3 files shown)


src/modules/Packages.rb
```diff
@@ -1741,6 +1741,10 @@ def Initialize_StageInitial(show_popup, base_url, log_url)
         end
       end
 
+      # Lower the priority of the initial installation repository to prefer
+      # the packages from the other media (bsc#1071742)
```
jreidinger: Does this cause in opensuse installation to prefer network over DVD when there are identical packages?

lslezak: Hmm... good point.

Then the DVD preference would not work correctly... :worried: 

We should do it only when *all* repos are DVD...

lslezak: Which means we cannot do it at this place as the addons are added later and we do not know in advance if all will be DVDs...

lslezak:
Closing, we need a different solution for this...

PR #310, opened 2018-01-11
Decrease the priority of the initial DVD installation repository (bsc#1071742)

lslezak:
To prefer the packages from the other DVD media, avoid media changes between the "Packages" and the "Installer" DVDs.

- See https://bugzilla.suse.com/show_bug.cgi?id=1071742
- Manually tested with the SLE15 "Installer" media + "Basesystem", "Server application" modules from the "Packages" medium. Everything was installed from the "Packages" medium, without any media change.
- It should work correctly with the openSUSE DVD + OSS online repositories, in that case the priority won't be changed and the DVD will be preferred (keeps the old behavior)
- See the Travis log for the [RSpec test cases output](https://travis-ci.org/yast/yast-packager/builds/327722678#L1298)
- 4.0.28

### DIFF
--- a/test/packages_test.rb
+++ b/test/packages_test.rb
@@ -1403,4 +1403,97 @@ describe Yast::Packages do
       expect { URI.parse(@base_url.value) }.to_not raise_error
     end
   end
+
+  describe "#adjust_repo_priority" do
+    before do
+      expect(Yast::Pkg).to receive(:SourceGetCurrent).with(true).and_return(repos.keys)
+      repos.each do |id, data|
+        expect(Yast::Pkg).to receive(:SourceGeneralData).with(id).and_return(data).at_least(:once)
+      end
+    end
+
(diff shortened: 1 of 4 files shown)


test/packages_test.rb
```diff
@@ -1403,4 +1403,97 @@ def product(properties = {})
...
+    context "installing from a DVD with 2 DVD addons" do
+      let(:repos) { { 0 => volatile_repo, 1 => volatile_repo, 2 => volatile_repo } }
+      it_behaves_like "sets priority", 0, 100
+    end
+
+    context "installing from a DVD with an USB and a DVD addon" do
+      let(:repos) { { 0 => volatile_repo, 1 => local_repo, 2 => volatile_repo } }
+      it_behaves_like "sets priority", 0, 100
+    end
+  end
```
jreidinger: nice behavior description.

PR #311, opened 2018-01-19
Fixed reading available products

lslezak:
- Ignore installed products at system upgrade
- Fully reset the packages status

### DIFF
--- a/src/lib/y2packager/product_reader.rb
+++ b/src/lib/y2packager/product_reader.rb
@@ -108,10 +108,13 @@ module Y2Packager
     def available_products
       products = Yast::Pkg.ResolvableProperties("", :product, "")
 
+      # remove e.g. installed products
+      products.select! { |p| p["status"] == :available || p["status"] == :selected }
+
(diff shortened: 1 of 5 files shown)


src/lib/y2packager/product_reader.rb
```diff
@@ -108,10 +108,13 @@ def product_package(name, repo_id)
...
       products = Yast::Pkg.ResolvableProperties("", :product, "")
 
+      # remove e.g. installed products
+      products.select! { |p| p["status"] == :available || p["status"] == :selected }
+
       # remove duplicates, there migth be different flavors ("DVD"/"POOL")
       # or archs (x86_64/i586), when selecting the product to install later
       # libzypp will select the correct arch automatically
-      products.uniq! { |prod| prod["name"] }
+      products.uniq! { |prod| "#{prod["name"]}__#{prod["version"]}" }
```
imobachgs: np: Why not use `p` as identifier, as you did 5 lines before?

PR #314, opened 2018-01-26
Fix mem consumtion hotfix

jreidinger:


src/modules/Packages.rb
```diff
@@ -2411,7 +2411,7 @@ def extra_system_patterns
     def log_software_selection
       log.info "-" * 10 << " Transaction Status Begin " << "-" * 10
 
-      RESOLVABLE_TYPES.each do |type|
+      [:product, :pattern].each do |type|
```
lslezak: NP: I'd add a comment explaining why we do not log packages. To avoid adding it back later....

PR #315, opened 2018-01-29
Reimplemented package reset to require less memory

lslezak:
- Related to [bsc#106768](https://bugzilla.suse.com/show_bug.cgi?id=1076768)
- 4.0.33

### DIFF
--- a/src/lib/y2packager/clients/inst_product_upgrade_license.rb
+++ b/src/lib/y2packager/clients/inst_product_upgrade_license.rb
@@ -71,11 +71,14 @@ module Y2Packager
       end
 
       # Reset packages having the specific status
-      # @param status [Symbol] Status symbol, `:removed`, `:selected`, `:installed`, `:available`
+      # @param status [Symbol] Status symbol, `:removed`, `:selected`, `:installed`,
+      # `:available`, `:taboo`, `:locked`
       # @see Yast::Pkg.GetPackages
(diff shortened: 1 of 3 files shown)


src/lib/y2packager/clients/inst_product_upgrade_license.rb
```diff
@@ -64,13 +64,19 @@ def product
...
 
         @product
       end
+
+      # Reset packages having the specific status
+      # @param status [Symbol] Status symbol, `:removed`, `:selected`, `:installed`, `:available`
+      # @see Yast::Pkg.GetPackages
+      def reset_packages(status)
+        # packages in the required status
+        Yast::Pkg.GetPackages(status, true).each { |p| Yast::Pkg.PkgNeutral(p) }
```
mvidner: What does this `true` mean?
Do we have an online API reference for `Pkg`?

mvidner: OK, this is it https://yast-pkg-bindings.surge.sh/classYast_1_1Pkg.html#a62a42363467a84326560adef5f795ec1

mvidner: For extra clarity, I would use
```rb
names_only = true
...GetPackages(status, names_only)...
```

PR #316, opened 2018-02-01
Y2packager move

lslezak:
- Related to https://github.com/yast/yast-yast2/pull/686


package/yast2-packager.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Feb  1 08:41:35 UTC 2018 - lslezak@suse.cz
+
+- Move some Y2Packager files to yast2 package to avoid circular
+  dependency (related to fate#1076513)
```
gilsonsouza: Fate number is probably wrong.

package/yast2-packager.changes
```diff
@@ -1,3 +1,10 @@
...
+- 4.0.34
```
gilsonsouza: The version was changed to 4.0.33 and not 34.

PR #322, opened 2018-02-13
Display the addon licenses from libzypp (bsc#1057223)

lslezak:
- [Trello card](https://trello.com/c/NFPbGpjO), [bsc#1057223](https://bugzilla.suse.com/show_bug.cgi?id=1057223)
- Added unit test
- Tested also manually using the latest build + updated `yast2-pkg-bindings` package

### Screenshot

Tested with the HA product from the Packages DVD:

addon_license_agreement2

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -1646,7 +1646,7 @@ module Yast
 
         locales.each do |locale|
           license = Pkg.PrdGetLicenseToConfirm(product_name, locale)
-          next unless license && !license.empty?
+          next if license.nil? || license.empty?
 
           found_license = true
           log.info("Found license for language #{locale.inspect}, size: #{license.bytesize} bytes")
@@ -1655,6 +1655,10 @@ module Yast
           File.write(path, license)
           log.info("License saved to #{path}")
         end
+
+        # we can display only one license at one time, there should be only one
+        # product per addon repository anyway
+        break if found_license
       end
 
       if found_license
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 require "yast"
 require "uri"
+require "fileutils"
 
 # Yast namespace
 module Yast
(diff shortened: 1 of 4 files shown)


src/modules/ProductLicense.rb
```diff
@@ -1182,19 +1183,18 @@ def SearchForLicense_NormalRunBaseProduct(_src_id, fallback_dir)
     def SearchForLicense_AddOnProduct(src_id, _fallback_dir)
       Builtins.y2milestone("Getting license beta from repository %1", src_id)
 
+      # use a separate license directory for each product
+      @tmpdir = File.join(SCR.Read(path(".target.tmpdir")), "product-license", src_id.to_s)
+
+      # try reading the product license from libzypp first
+      return if product_license(src_id, @tmpdir)
```
jreidinger: it will skip reading `@beta_file`, is it ok?

src/modules/ProductLicense.rb
```diff
@@ -1624,6 +1624,59 @@ def beta_seen?(id)
...
+          log.info("License saved to #{path}")
+        end
+      end
+
+      if found_license
+        @license_dir = tmpdir
+        @license_file_print = repo_license_file(id)
+      end
+
+      found_license
```
jreidinger: this means if we found license for at least one product, we are done with licenses?

src/modules/ProductLicense.rb
lslezak: Yes, that's OK. The Beta file cannot be placed into RPM-MD repositories, it's not accessible via libzypp from there.

src/modules/ProductLicense.rb
```diff
@@ -1624,6 +1624,59 @@ def beta_seen?(id)
...
+      log.info("Found products from source #{id}: #{product_names.inspect}")
+      found_license = false
+
+      product_names.each do |product_name|
+        locales = Pkg.PrdLicenseLocales(product_name)
+        log.info("License locales for product #{product_name.inspect}: #{locales.inspect}")
+
+        locales.each do |locale|
+          license = Pkg.PrdGetLicenseToConfirm(product_name, locale)
+          next unless license && !license.empty?
```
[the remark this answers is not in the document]
lslezak: Yeah, probably converted by Rubocop, I'll check it...

src/modules/ProductLicense.rb
lslezak: Yes, we can display license only for one product anyway. That's compatible with the old approach, there could be only one `license.tar.gz` file on the medium. No regression.

PR #326, opened 2018-02-26
Install the packages needed for accessing the repositories (bsc#1063980)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1063980
- The `InstMode` value has been dropped from `install.inf`, not written by Linuxrc anymore
- Instead evaluate the repository URL
- Additionally evaluate also the add-on repositories, not just the base product repository
- Install e.g. `nfs-client` for `nfs://` repositories
- Properly evaluate the `iso://` repositories (the real URL is set in the `url` query parameter)
- 4.0.42

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -5,6 +5,7 @@ require "yast"
 require "erb"
 require "fileutils"
 require "uri"
+require "cgi"
 
 # Yast namespace
 module Yast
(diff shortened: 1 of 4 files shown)


src/modules/Packages.rb
```diff
@@ -1047,26 +1048,57 @@ def boardPackages
...
+    # Return the URL schemes for the currently defined repositories (only enabled
+    # repositories are evaluated). For ISO repositories the base URL schema
+    # is returned (i.e. the location of the ISO), invalid URLs are ignored.
+    # @return [Array<String>] the list of the URL schemes, empty if
+    #   no repository is defined
+    def repo_schemes
+      schemes = []
+
+      # all enabled repositories
+      Pkg.SourceGetCurrent(true).map do |repo|
```
imobachgs: As you are not doing anything with the return value, you could use a plain `each` or, if you prefer, a `each_with_object`.

test/packages_test.rb
```diff
@@ -36,7 +36,8 @@ def load_zypp(file_name)
 
 PRODUCTS_FROM_ZYPP = load_zypp("products.yml").freeze
 
-describe Yast::Packages do
+describe "Yast::Packages" do
```
imobachgs: Why do you need to use the string form?

lslezak: Just because of a better formatting.

It's not strictly needed, but as `Yast::Packages` is an object (not a class) then RSpec prints 
`#<Yast::PackagesClass:0x00000000028047b8>` in the output (see [here](https://travis-ci.org/yast/yast-packager/builds/345128635#L1107)). That does not look nice, printing `Yast::Packages` looks better IMHO.

src/modules/Packages.rb
lslezak: Oh, originally I started the code with `map`, but it turned out that there is no mapping for some (invalid) values and I'd like to avoid duplicates. So in the end I used a different solution.

PR #331, opened 2018-03-09
Allow next back

jreidinger:
https://trello.com/c/WvlPBpFV/1356-1-upgrade-make-sure-that-packages-dvd-and-other-add-ons-can-be-used-from-summary-screen

tested manually


lslezak:
At the daily call you said that the client is called with `false, false` arguments, shouldn't we rather change the caller code to pass `true, true`?

PR #338, opened 2018-03-28
Use the new license API.

teclator:
- It depends on yast/yast-yast2#715

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -3,6 +3,9 @@ require "yast"
 require "uri"
 require "fileutils"
 
+require "y2packager/product"
+require "y2packager/product_license"
+
 # Yast namespace
 module Yast
   # Provide access / dialog for product license
@@ -81,7 +84,7 @@ module Yast
       @beta_file_already_seen = {}
     end
 
-    # Returns whether accepting the license manually is requied.
+    # Returns whether accepting the license manually is required.
     #
     # @see BNC #448598
     # @param [Any] id unique ID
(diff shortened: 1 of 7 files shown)


src/modules/ProductLicense.rb
```diff
@@ -778,6 +785,44 @@ def display_beta(id)
...
+    #
+    # @param product  [Y2Packager::Product] Product instance
+    # @param licenses [Hash<String,String>] License files indexed by language
+    # @return [boolean] true if the license was already accepted; false if it was
+    #   not acceptedd or it is not found.
+    def license_accepted_for?(product, licenses)
+      license_file = licenses["en_US"] || licenses["en"] || licenses[""]
+      content = SCR.Read(path(".target.string"), license_file).to_s
+      if content.empty?
+        log.info "No license was found for #{product.name} in the repository"
```
jreidinger: are you sure it is just info? and do you test how it behave for this case?

src/modules/ProductLicense.rb
```diff
@@ -778,6 +785,44 @@ def display_beta(id)
...
+    # Find the product in the given repository
+    #
+    # @param src_id [Integer] Repository ID
+    # @return [Y2Packager::Product,nil] Product or nil if it was not found
+    def repository_product(src_id)
+      product_h = Yast::Pkg.ResolvableProperties("", :product, "").find do |properties|
+        properties["source"] == src_id
+      end
+      if product_h.nil?
+        log.info "No product was found in the repository (#{src_id})" unless product_h
```
jreidinger: can this happen at all?

jreidinger: and useless `unless`

imobachgs: Not for official addons, but I prefer to be on the safe side.

src/modules/ProductLicense.rb
imobachgs: Unit tests should cover this case. I will raise it to error.

PR #341, opened 2018-04-10
bnc 1086846 - selectin add-on products

schubi2:


test/addon_product_test.rb
```diff
@@ -290,6 +290,39 @@
...
+      let(:repo_id) { 1 }
+      let(:repo) do
+        ADDON_REPO.merge("install_products" => ["available_product",
+                                                "not_available_product"])
+      end
+
+      before do
+        allow(subject).to receive(:AcceptedLicenseAndInfoFile).and_return(true)
+        allow(Yast::Pkg).to receive(:SourceProductData).with(repo_id)
+        expect(Yast::Pkg).to receive(:ResolvableInstall).with("available_product",
```
imobachgs: You should not call `expect` in the `before` block (you should use it to set up the test). Changing to `allow` should do the trick.

If you want to explicitly test this call, set this one to `allow` and move the `expect` to the corresponding `it` block. For instance to something like:

```ruby
it "tries to install given products" do
  expect(Yast::Pkg).to receive(:ResolvableInstall).with("available_product", :product).and_return(true)
  expect(Yast::Pkg).to receive(:ResolvableInstall).with("not_available_product", :product).and_return(false)
  subject.AddPreselectedAddOnProducts(filelist)
end
```

schubi2:
Ah, nice ! I have not been aware that expects can "overrule" the allows.

PR #342, opened 2018-04-12
bsc#1078323: Warning msg for a repo that belongs to a service.

gilsonsouza:
warning_repo_service

### DIFF
--- a/src/clients/repositories.rb
+++ b/src/clients/repositories.rb
@@ -80,6 +80,10 @@ module Yast
         "guihandler" => fun_ref(method(:StartInstSource), "symbol ()")
       }
 
+      # list of repositories that was already informed to the user that they are
+      # managed by service
+      @services_repos = []
+
       if WFM.Args == [:sw_single_mode]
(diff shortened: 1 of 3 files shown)


src/clients/repositories.rb
```diff
@@ -1939,6 +1944,16 @@ def StartInstSource
       UI.CloseDialog
       ret
     end
+
+    def warn_service_repository(source_state)
+      msg = _("Repo %{name} is managed by service %{service}.\n"\
+        "Volatile changes are reset by the next service refresh!") % {name: source_state["name"],
```
lslezak: I'd rephrase this part of the message "Your manual changes will be reset by the next service refresh!"

lslezak:
BTW for the customer it would be enough even for fixing it in SP4, so I think we should skip SP2 and fix it in SP3+ only. (This is not a critical fix and that's how libzypp behaves for ages, no regression.)

See https://bugzilla.suse.com/show_bug.cgi?id=1078323#c37

PR #348, opened 2018-04-25
Fix "installation settings" support for RTL languages

imobachgs:
* Related to https://github.com/yast/yast-yast2/pull/730 and https://github.com/yast/yast-installation/pull/687.

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -392,7 +392,8 @@ module Yast
         if Ops.greater_than(Builtins.size(patterns), 0)
           output = Builtins.add(
             output,
-            Ops.add(_("Patterns:"), HTML.List(patterns))
+            # TODO: simply delete the br once we can afford to break translations again
+            Ops.add(_("Patterns:<br>").chomp("<br>"), HTML.List(patterns))
           )
         end
       end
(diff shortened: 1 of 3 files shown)


src/modules/Packages.rb
```diff
@@ -387,12 +387,12 @@ def SummaryOutput(flags)
       end
 
       if Builtins.contains(flags, :pattern)
-        patterns = ListSelected(:pattern, "+  %1")
+        patterns = ListSelected(:pattern, "%1")
 
         if Ops.greater_than(Builtins.size(patterns), 0)
           output = Builtins.add(
             output,
-            Ops.add(_("Patterns:<br>"), Builtins.mergestring(patterns, "<br>"))
```
mvidner: I am not sure whether it's better to remove a line break than keep it translated. Maybe it will look acceptable even with the br?

imobachgs: I was unsure too. However, with the `br` looks weird to me (although it is acceptable):

installer-settings-weird-spacing-detail

I was considering adding a br after the patterns list, but it looks like a waste of space.

src/modules/Packages.rb
```diff
@@ -387,12 +387,13 @@ def SummaryOutput(flags)
...
+            # TODO: simply delete the br once we can afford to break translations again
+            Ops.add(_("Patterns:<br>").chomp("<br>"), HTML.List(patterns))
```
mvidner: Yay, I think we can have the best of both solutions, with a small hack™.

imobachgs:
Nice solution :)

PR #355, opened 2018-05-29
Fix licenses display in textmode

imobachgs:
When running the installer on textmode, the license translation will be limited to the language that was selected when booting. If the translation is not available for that language, the english version will be displayed.

The graphical interface is not affected by this change.

### textmode=1 lang=ja_JP

ja-jp-translated

### textmode=1 (selecting Japanese on the welcome screen)

ja-jp-not-translated

### textmode=0 (selecting Japanese on the welcome screen)

ja-jp-translated-gui

### textmode=0 (selecting Slovak -no translated- on the welcome screen)

slovak-missing-translation


src/lib/y2packager/widgets/product_license_translations.rb
```diff
@@ -68,15 +71,53 @@ def handle(event)
...
       # @return [Widgets::ProductLicenseContent]
       def product_license
         @product_license ||=
-          Y2Packager::Widgets::ProductLicenseContent.new(product, language)
+          Y2Packager::Widgets::ProductLicenseContent.new(product, content_language)
+      end
+
+      # Available license translations
+      #
+      # When running on textmode, only the preselected/given language is considered.
```
lslezak: Please also mention *why* we need to handle the text mode differently.

src/lib/y2packager/widgets/product_license_translations.rb
```diff
@@ -68,15 +71,61 @@ def handle(event)
...
+      # see #default_language for further details.
+      #
+      # @return [Array<String>] Locale codes of the available translations
+      # @see #default_language
+      def available_locales
+        return product.license_locales unless Yast::UI.TextMode
```
lslezak: Maybe for the consistency with the method below use the triple operator here as well:
```ruby
Yast::UI.TextMode ? [default_language] : product.license_locales
```

this also avoid `unless` which is often confusing...

PR #361, opened 2018-06-12
Sort names

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=957372

before:
repos_before

after:
repos_after

### DIFF
--- a/test/lib/clients/repositories_test.rb
+++ b/test/lib/clients/repositories_test.rb
@@ -105,4 +105,46 @@ describe Yast::RepositoriesClient do
       end
     end
   end
+
+  describe "#SortReposByPriority" do
+    it "returns nil if param is nil" do
+      expect(client.SortReposByPriority(nil)).to eq nil
+    end
+
+    it "returns input sort by priorities" do
+      repos = [
+        { "priority" => 10, "name" => "repo1" },
+        { "priority" => 30, "name" => "repo2" },
+        { "priority" => 20, "name" => "repo3" }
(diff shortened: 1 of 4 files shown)


test/lib/clients/repositories_test.rb
```diff
@@ -105,4 +105,46 @@
       end
     end
   end
+
+  describe "#SortReposByPriority" do
+    it "returns nil if param is nil" do
+      expect(client.SortReposByPriority(nil)).to eq nil
```
lslezak: NP: I'd slightly prefer `be_nil` instead of `eq nil` as it sounds more like spoken English, but it depends on your taste...

test/lib/clients/repositories_test.rb
```diff
@@ -105,4 +105,46 @@
...
+    end
+
+    it "returns input sort by priorities" do
```
lslezak: Use `it "sorts by priorities"`, sounds better and is consistent with the other test below.

test/lib/clients/repositories_test.rb
jreidinger: well, I do not like be_nil as it makes test less explicit. I want to ensure that it really is nil and not some fancy object, that return true for `nil?`

lslezak: Redefining `#nil?` to return `true` would be one of the dirtiest hacks ever. The code which does that is doomed.

PR #364, opened 2018-07-16
Fixed free space overflow for disks > 8EiB (bsc#991090)

lslezak:
Fixes a bug where a "not enough free space" popup is displayed for disks where the free space is bigger than 8EiB (2<sup>63</sup>). The problem is that `YCPInteger` internally uses signed 64-bit `long long` and numbers bigger than 2<sup>63</sup> become negative.

See [bug#1073696](https://bugzilla.suse.com/show_bug.cgi?id=991090) for more details.

not_enough_space

So if we get a negative free space then we skip the check, more than 8EiB free space should be always enough so far. :smile: 

## Note

The sizes reported by libzypp are in KiB, that means the libzypp API can support sizes >8EiB without causing data overflow (the limit is 8ZiB). The problem is with the `Pkg.TargetAvailable` call which returns the size in bytes and is affected by the `YCPInteger` limit.

## Additional Fixes

- The "not enough free space" warning is now skipped when there is nothing to install to that partition. Originally the code complained even for full data partitions where nothing was going to install. See [bug#926841](https://bugzilla.suse.com/show_bug.cgi?id=926841).
- When comparing the code with `master` I found yet another fix which is worth of backporting to SP3  (and SP4) - that's the check for non-existing directories. See [bug#1073696](https://bugzilla.suse.com/show_bug.cgi?id=1073696) an PR #307.

### DIFF
--- a/src/modules/SlideShowCallbacks.rb
+++ b/src/modules/SlideShowCallbacks.rb
@@ -10,9 +10,12 @@
 # $Id$
 #
 require "yast"
+require "pathname"
 
 module Yast
   class SlideShowCallbacksClass < Module
+    include Yast::Logger
+
     def main
       Yast.import "Pkg"
       Yast.import "UI"
@@ -300,8 +303,13 @@ module Yast
         if pkgdu != nil
           # check each mount point
           Builtins.foreach(pkgdu) do |part, data|
+            # disk sizes from libzypp, in KiB!
+            _disk_size, used_now, used_future, read_only = *data
+            # the size difference, how much the package needs on this partition
+            required_space = used_future - used_now
+
             # skip read-only partitions, the package cannot be installed anyway
-            if Ops.get(data, 3, 0) == 1
+            if read_only == 1
               Builtins.y2debug("Skipping read-only partition %1", part)
               next
             end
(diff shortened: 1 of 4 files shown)


src/modules/SlideShowCallbacks.rb
```diff
@@ -310,21 +318,47 @@ def DisplayStartInstall(pkg_name, pkg_location, pkg_description, pkg_size, delet
...
+
+            # nothing to install on this partition, skip it (bsc#926841)
+            if required_space <= 0
+              log.debug("Nothing to install at #{part}, skipping")
+              next
+            end
+
+            target_dir = File.join(Installation.destdir, part)
+
+            # handle missing directories (not existing yet or incorrect metadata),
```
shundhammer: Not sure if this can happen, but please consider this scenario:

During installation, the target filesystems are mounted below /mnt; if any separate filesystem on the target could not be mounted, so that directory isn't there, will it move up the hierarchy to / which is the inst-sys root filesystem which is read-only or has very little available space. Can this cause a problem here?

lslezak: If the target directory below `/mnt` is not there then it should stop at `/mnt`. If `/mnt` is not there than a wrong disk space check would be the least problem you had, the system would be so broken that you cannot continue anyway...

lslezak:
It's not a real problem, that situation should never happen. And if it does then the incorrect space check is just a minor issue as the installation is already completely messed up...

PR #367, opened 2018-07-18
Merge the 8EiB fix from SLE12 to SLE15

lslezak:
Merging #364 to SLE15.

And it seems there was added some extra unit test in `test/packages_test.rb` to SLE12, merging it as well...


shundhammer:
LGTM

PR #374, opened 2018-09-13
Improved self-update to use some packages as an add-on (bsc#1101016)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1101016

## Notes

The full functionality depends also on these other pull requests:
- https://github.com/yast/yast-yast2/pull/826
- https://github.com/yast/yast-installation/pull/741

## Tested Manually

- Tested manually with a customized `skelcd-control-SLES` package (Knut, thanks! :wink: ), applied using a DUD with `self_update=` boot parameter pointing to a repository with the new skelcd.
- Also tested an AutoYaST installation
- The extra addon repository is removed at the end

self-update-role

### DIFF
--- a/src/lib/y2packager/clients/inst_repositories_initialization.rb
+++ b/src/lib/y2packager/clients/inst_repositories_initialization.rb
@@ -15,7 +15,7 @@
 
 require "yast"
 require "y2packager/product"
-require "y2packager/selfupdate_addon_repo"
+require "y2packager/self_update_addon_repo"
 
 Yast.import "Packages"
 Yast.import "PackageCallbacks"
@@ -71,8 +71,8 @@ module Y2Packager
         Yast::PackageCallbacks.RestorePreviousProgressCallbacks
 
         # add extra addon repo built from the initial self update repository (bsc#1101016)
-        if Y2Packager::SelfupdateAddonRepo.present?
-          Y2Packager::SelfupdateAddonRepo.create_repo
+        if Y2Packager::SelfUpdateAddonRepo.present?
+          Y2Packager::SelfUpdateAddonRepo.create_repo
         end
 
         true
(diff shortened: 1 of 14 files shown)


src/lib/y2packager/selfupdate_addon_repo.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def self.present?(path = REPO_PATH)
+      # the directory exists and is not empty
+      ret = File.exist?(path) && !Dir.empty?(path)
+      log.info("Repository #{path} exists: #{ret}")
+      ret
+    end
+
+    def self.create_repo(path = REPO_PATH)
+      # create a plaindir repository, there is no package metadata
```
teclator: What about moving this comment to method documentation

src/lib/y2packager/selfupdate_addon_repo.rb
```diff
@@ -0,0 +1,65 @@
...
+        downloader = Packages::PackageDownloader.new(repo_id, pkg)
+        log.info("Downloading package #{pkg}...")
+        downloader.download(File.join(path, "#{pkg}.rpm"))
+      end
+
+      log.info("Downloaded packages: #{Dir["#{path}/*"]}")
+
+      true
+    end
+
```
teclator: It is almost self explanatory and tested, having a method documentation will make it perfect ;)

src/lib/y2packager/clients/inst_repositories_initialization.rb
```diff
@@ -68,6 +69,12 @@ def init_installation_repositories
         # initialize libzypp and get the base product name (intentionally not translated)
         # FIXME: UI.SetProductName(Product.name || "SUSE Linux")
         Yast::PackageCallbacks.RestorePreviousProgressCallbacks
+
+        # add extra addon repo built from the initial self update repository (bsc#1101016)
+        if Y2Packager::SelfupdateAddonRepo.present?
```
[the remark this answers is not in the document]
lslezak: Um, that would end at column 96... that would still fit but as this is a condition and the `if` is better visible this way.

PR #375, opened 2018-09-21
tag: confirm_license added in add_on_products.xml

schubi2:


src/modules/AddOnProduct.rb
```diff
@@ -1463,6 +1463,15 @@ def AddPreselectedAddOnProducts(filelist)
           priority = one_product.fetch("priority", -1).to_i
           prodname = one_product.fetch("name", "")
           check_name = one_product.fetch("check_name", true)
+          confirm_license = if Mode.auto
```
lslezak: You can simplify it to an one liner :wink:

```ruby
# the default value in AutoYaST is false, otherwise it is true
confirm_license = one_product.fetch("confirm_license", !Mode.auto)
```

doc/license_code_cleanup.md
```diff
@@ -146,6 +146,17 @@ the licenses acceptance completely (reading, showing and accepting license)
...
 
+4.0 Adding products via file add_on_products.xml
+==================================================
+
+While installation it is possible to add products automatically. The concerning product
+has already to be on the medium. Adding products will be defined by the file
+add_on_products.xml in the root directory of the installation source.
+The flag -confirm_license- defines if the user has to accepted the license which has been
+defined in the product.
+
```
lslezak: There is one more use case for the `add_on_products.xml` file, maybe it would make sense to add this note:

```
Note: The `add_on_products.xml` file is also read from the installation system (inst-sys)
root directory (/). It is created by linuxrc when using a driver update. In that case YaST
automatically prefers the updated packages from the driver update instead of the original
package from the media. This XML is for internal use by YaST and should not be changed
by users.
```

doc/license_code_cleanup.md
```diff
@@ -146,6 +146,17 @@ the licenses acceptance completely (reading, showing and accepting license)
 It can handle all license types ( license is in /license.tar.gz, SCC license and libzypp license )
 License types are stored under /tmp/YaST2-<number>/product-license .
 
+4.0 Adding products via file add_on_products.xml
+==================================================
+
+While installation it is possible to add products automatically. The concerning product
+has already to be on the medium. Adding products will be defined by the file
+add_on_products.xml in the root directory of the installation source.
+The flag -confirm_license- defines if the user has to accepted the license which has been
```
lslezak: Typo: "user has to accept the license"

PR #385, opened 2018-11-14
Add online repos during installation

shundhammer:
## Trello

https://trello.com/c/VgNO1EBi/442-ludwigs-favorite-feature-leap-151-add-possibility-to-add-online-updates-repo-during-installation

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1112937

## Changes Summary

- Support for adding online repos during installation 

- Added pop-up dialog to ask user first if he would like to add online repos during installation and upgrade (only once)

## Details

During installation or upgrade, this now opens a pop-up dialog first:

yast2-001

If memory is low, it also comes with a warning:

yast2-001

(only one of those two is displayed)

If the user clicks on "Yes", the next step is configuring which online repos to use:

yast2-002

When going back and forth in the workflow, the behavior is consistent:

- The pop-up is only posted once if the user answered with "Yes", so the full-fledged dialog is then shown immediately, no matter how often the user goes back and forth.

- If the user answered the pop-up with "No", the module is skipped when going backwards, so the first dialog (license / language / keyboard) is shown; when going forward again, the pop-up is shown again so the user still has a chance to get online repos.

In other modes, there may be the old low memory warning dialog. The module now makes sure to show that warning only once; if the new pop-up is shown, it already contains that warning, so the old warning is not shown.

## Related PR

https://github.com/yast/skelcd-control-openSUSE/pull/151

## Stakeholder Feedback

Demonstrated to and approved by @lnussel.

He extended the first (simpler) pop-up dialog text, and he requested to skip this workflow step entirely if no network is available.

His reasoning was that the network is configured early in the workflow anyway, and having yet another network configuration from this online repos module (which it can do!) would only add confusion rather than be helpful.

### DIFF
--- a/src/lib/y2packager/clients/inst_productsources.rb
+++ b/src/lib/y2packager/clients/inst_productsources.rb
@@ -87,6 +87,7 @@ module Yast
   #     </group>
   #   </metapackage>
   class InstProductsourcesClient < Client
+    # rubocop:disable Style/ClassVars
     include Yast::Logger
 
     # too low memory for using online repositories (in MiB),
@@ -119,6 +120,7 @@ module Yast
         end
 
         if !ask_activate_online_repos
+          log.info("Skipping module")
           return :auto
         end
       end
--- a/src/lib/y2packager/clients/inst_productsources.rb
+++ b/src/lib/y2packager/clients/inst_productsources.rb
@@ -1739,6 +1739,7 @@ module Yast
       # Ask only once
       return @@ask_activate_online_repos_result unless @@ask_activate_online_repos_result.nil?
 
+      default_button = :focus_yes
       msg = _("Your system has an active network connection.\n" \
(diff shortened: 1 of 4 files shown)


src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -1681,17 +1697,69 @@ def check_memory_size
       return if !Mode.installation
       # less than LOW_MEMORY_MIB RAM, the 64MiB buffer is for possible
       # rounding in hwinfo memory detection (bsc#1045915)
```
lslezak: Please move this comment also below, the "64MiB buffer" part belongs to that `- 64` subtraction.

shundhammer: Ah, right. Moved.

PR #389, opened 2018-11-19
Disable cd / dvd repositories when installation is over

mchf:
Fate#325541

Depends on https://github.com/yast/yast-installation-control/pull/72

### DIFF
--- a/src/lib/packager/clients/pkg_finish.rb
+++ b/src/lib/packager/clients/pkg_finish.rb
@@ -102,6 +102,19 @@ module Yast
 
   private
 
+    # Checks if given source repo should be disabled
+    #
+    # Currently CD / DVD repositories are disabled after installation, when
+    # enabled in control file
+    #
+    # @param source [Symbol] if :cd or :dvd then control file is checked if the
+    #                        repo should be disabled
+    # @return [Boolean] true if the repo should be disabled
+    def disable_media_repo?(source)
+      disable_repo = ProductFeatures.GetBooleanFeature("software", "disable_media_repo")
+      [:cd, :dvd].include?(source) && disable_repo
+    end
+
     # Backup old sources
     #
     # During upgrade, old sources are reinitialized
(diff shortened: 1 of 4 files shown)


test/pkg_finish_test.rb
```diff
@@ -77,19 +80,24 @@
...
 
     context "given some local repository" do
-      let(:repositories) { [local_repo, remote_repo] }
+      let(:repositories) { [local_repo, local_dvd_repo, remote_repo] }
       let(:base_products) { [sles_product, sled_product] }
 
       let(:local_repo) do
         Y2Packager::Repository.new(repo_id: 1, name: "SLE-12-SP2-0", enabled: true,
-          url: URI("cd://dev/sr0"), autorefresh: false)
+          url: URI("dir:///path/to/extracted_ISO"), autorefresh: false)
```
lslezak: The `dir://` schema makes only sense in installed system, you cannot use it for installing a system (`/` is in inst-sys in RAM disk).

But you can use the `hd`or `iso` schema for installation, like `hd:/dir?device=/dev/sda1`. For details see "Supported URI formats" section in `man zypper`.

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -102,6 +102,19 @@ def write
...
+    # Checks if given source repo should be disabled
+    #
+    # Currently CD / DVD repositories are disabled after installation, when
+    # enabled in control file
+    #
+    # @param source [Symbol] if :cd or :dvd then control file is checked if the
+    #                        repo should be disabled
+    # @return [Boolean] true if the repo should be disabled
+    def disable_media_repo?(source)
+      disable_repo = ProductFeatures.GetBooleanFeature("software", "disable_media_repo")
```
lslezak: Just an optimization, we do not have to evaluate the `disable_media_repo` option if the source is not CD or DVD.

```ruby
return false unless [:cd, :dvd].include?(source)
ProductFeatures.GetBooleanFeature("software", "disable_media_repo")
```

test/pkg_finish_test.rb
mchf: well, I picked the first one which came to my mind and was not cd/dvd ... but you're right, I'll do it more "real world" example

src/lib/packager/clients/pkg_finish.rb
mchf: ok

PR #390, opened 2018-11-19
Report chrony is missing

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1082369

## Warn the user when NTP was configured but removed from the software selection in the proposal.

chronymissing

Note: Do not like adding the dependency to yast-packager just for the warning.

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -80,6 +80,7 @@ module Yast
       Yast.import "URL"
       Yast.import "PackagesProposal"
       Yast.import "HTML"
+      Yast.import "NtpClient"
 
       Yast.include self, "packager/load_release_notes.rb"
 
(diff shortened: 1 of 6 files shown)


src/modules/Packages.rb
```diff
@@ -857,6 +858,18 @@ def InitializeAddOnProducts
...
+    # When the NTP configuration has been modified, it checks whether the
+    # service package is selected to be installed or not.
+    # @return [String] empty string or error message if the package is missing
+    def check_ntp_installation_packages
+      return "" unless NtpClient.modified
+      return "" if pkg_will_be_installed(NtpClientClass::REQUIRED_PACKAGE)
+
+      # TRANSLATORS: warning message, %s is the service name
+      _("The NTP configuration (%s) has been modified, " \
+        "but the package is not selected to be installed.") % "chrony"
```
jreidinger: not sure if that chrony should be or not translated. Eg. consider chinese.

src/modules/Packages.rb
```diff
@@ -80,6 +80,7 @@ def main
       Yast.import "URL"
       Yast.import "PackagesProposal"
       Yast.import "HTML"
+      Yast.import "NtpClient"
```
jreidinger: this adds dependency on ntp client always, which looks wrong. I think you should import it in method itself, so it will be imported only when needed ( during installation ).

PR #395, opened 2018-12-05
remove cd:/// fallback installation URL (fate#325482)

wfeldt:


test/inst_url_test.rb
```diff
@@ -29,8 +29,8 @@
     context "when ZyppRepoURL is not defined" do
       let(:zypp_repo_url) { nil }
 
-      it "returns cd:///" do
-        expect(inst_url.installInf2Url("")).to eq("cd:///")
+      it "returns an empty string" do
+        expect(inst_url.installInf2Url("")).to be_empty
```
lslezak: NP: The `be_empty` check is a bit tricky, it actually calls the `empty?` method internally which is also defined for arrays so `expect([]).to be_empty` also passes the test. So the "returns an empty string" statement is not completely true, the check would also pass when returning an empty list.

Yes, this is a corner case, but I think the `eq("")` would be a bit better here.

```suggestion
        expect(inst_url.installInf2Url("")).to eq("")
```

PR #396, opened 2018-12-14
Hardening

jreidinger:


src/clients/inst_kickoff.rb
```diff
@@ -44,8 +44,8 @@ def main
           # When upgrading system, remove devs.rpm if installed
```
lslezak: Hm, this has been added 12 years ago, the `devs` package has been dropped long time ago. https://maintenance.suse.de/maintained/?package=devs displays SLE10.

So I'd rather completely remove this obsolete code.

src/clients/inst_kickoff.rb
```diff
@@ -474,17 +429,9 @@ def createmdadm
...
-            Ops.add(
-              Ops.add("chroot '", String.Quote(Installation.destdir)),
-              "' "
-            ),
-            "mdadm -Ds"
-          )
-        )
+      out = SCR.Execute(
+        path(".target.bash_output"),
+        "/usr/bin/chroot #{Installation.destdir.shellescape} mdadm -Ds"
```
lslezak: Do we still need this? Maybe it's already handled by the `libstorage-ng`... Again, this code is ~13 years old... :open_mouth:

src/modules/Packages.rb
```diff
@@ -1227,7 +1228,7 @@ def ComputeSystemPackageList
     # Import GPG keys found in the inst-sys
     def ImportGPGKeys
       out = Convert.to_map(
-        SCR.Execute(path(".target.bash_output"), "/bin/ls -d /*.gpg")
+        SCR.Execute(path(".target.bash_output"), "/usr/bin/ls -d /*.gpg")
```
lslezak: NP: maybe we could use some native Ruby here, as it reads from inst-sys we do not need any chroot magic.

src/modules/Packages.rb
```diff
@@ -1702,11 +1683,7 @@ def FindAndRememberAddOnProductsFiles(initial_repository)
           )
           WFM.Execute(
             path(".local.bash"),
-            Builtins.sformat(
-              "cp '%1' '%2'",
-              String.Quote(tmp_add_on_products),
-              String.Quote(filename)
-            )
+            "/usr/bin/cp #{tmp_add_on_products.shellescape} #{filename.shellescape}"
```
lslezak: NP: Use [::FileUtils.cp](https://ruby-doc.org/stdlib-2.5.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-cp) here?

src/modules/Packages.rb
```diff
@@ -1367,18 +1360,12 @@ def FindAndCopySlideDir(our_slidedir, source, search_for_dir, lang_long,
...
 
-        copy_command = Builtins.sformat(
-          "cp -r '%1/%2/pic'/* '%3'",
-          String.Quote(imagesdir),
-          String.Quote(search_for_dir),
-          String.Quote(our_imagesdir)
-        )
+        source_path = File.join(imagesdir, search_for_dir, "pic", "*")
+
+        copy_command = "/usr/bin/cp -r #{source_path.shellescape} #{our_imagesdir.shellescape}"
```
lslezak: NP: Use [::FileUtils.cp_r](https://ruby-doc.org/stdlib-2.5.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-cp_r) here?

src/modules/SourceDialogs.rb
```diff
@@ -1966,11 +1966,10 @@ def ServerWidget
     def IsAnyNetworkAvailable
```
lslezak: We should replace the implementation by just calling [NetworkService.isNetworkRunning](https://github.com/yast/yast-yast2/blob/master/library/network/src/modules/NetworkService.rb#L280) or even drop it and update the callers to use it instead...

src/modules/SourceManager.rb
```diff
@@ -751,14 +753,9 @@ def InstInitSourceMoveDownloadArea
...
-        SCR.Execute(
-          path(".target.bash"),
-          Builtins.sformat(
-            "test -d '%1' || mkdir -p '%1'",
-            String.Quote(download_dir)
-          )
-        )
+      successful = SCR.Execute(
+        path(".target.bash"),
+        "test -d #{download_dir.shellescape} || /usr/bin/mkdir -p #{download_dir.shellescape}"
```
lslezak: I guess we could drop that `test` as `mkdir -p` does not fail if it already exists.

src/modules/Packages.rb
jreidinger: well, I do not where it is trivial enough, as I do not want to break it too much.

src/clients/inst_kickoff.rb
jreidinger: I do not know. @ancorgs ^^^

src/clients/inst_kickoff.rb
jreidinger: ok, I do not know it.

src/modules/Packages.rb
jreidinger: if we are sure that it cannot fail, then it is ok for me to change it.

src/modules/SourceDialogs.rb
jreidinger: ok

src/modules/SourceManager.rb
jreidinger: fine for me.

src/modules/Packages.rb
jreidinger: well, this won't work with fileutils as it uses "*" expansion in path. So I keep it as it is.

src/clients/inst_kickoff.rb
```diff
@@ -34,21 +34,7 @@ def main
...
-        else
-          # When upgrading system, remove devs.rpm if installed
-          LocalCommand(
-            Builtins.sformat(
-              "/bin/rpm --root %1 -q 'devs' && /bin/rpm --nodeps --root %1 -e 'devs'",
-              Installation.destdir.shellescape
-            )
-          )
-
+        if Installation.destdir != "/"
```
lslezak: NP: Merge these two conditions? `if Mode.update && Installation.destdir != "/"`

jreidinger: I am still not sure if we do not have more robust way to testing online upgrade. But fine for me to merge it.

PR #399, opened 2018-12-27
Display the license confirmation checkbox only if needed

dgdavid:
Related to https://trello.com/c/ish1gWEZ/568-part-of-fate325482-display-the-license-contained-in-the-fallback-repo (fate#325482)

---

The license confirmation can be ignored when the package contains the file/flag [`no-acceptance-needed`](https://github.com/openSUSE/libzypp/blob/5d59a317e7a8d523e569063abaaaf12c0aeab5bd/zypp/RepoInfo.cc#L757). In consequence, the confirmation checkbox only must be displayed when the acceptance is required.

---

See also https://github.com/yast/yast-yast2/pull/880

### DIFF
--- a/src/lib/y2packager/dialogs/inst_product_license.rb
+++ b/src/lib/y2packager/dialogs/inst_product_license.rb
@@ -55,10 +55,23 @@ module Y2Packager
         VBox(
           Widgets::ProductLicenseTranslations.new(product, Yast::Language.language),
           HBox(
-            Left(Widgets::ProductLicenseConfirmation.new(product)),
+            confirmation_checkbox,
             HStretch()
           )
         )
+      end
+
+    private
+
+      # Return the license confirmation widget if required
+      #
+      # It returns Empty() if confirmation is not needed.
+      #
+      # @return [Yast::Term,ProductLicenseConfirmation] Product confirmation license widget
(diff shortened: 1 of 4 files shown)


dgdavid:
#### Installation without required license acceptance

 Installing single product from fallback repo

src/lib/y2packager/dialogs/inst_product_license.rb
```diff
@@ -55,11 +55,24 @@ def contents
         VBox(
           Widgets::ProductLicenseTranslations.new(product, Yast::Language.language),
           HBox(
-            Left(Widgets::ProductLicenseConfirmation.new(product)),
+            confirmation_checkbox,
```
jreidinger: how it works previously for opensuse which does not have license or for some addons?

dgdavid: That's a good question.

Maybe because (the installer workflow?) is using https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/widgets/product_license.rb instead?

What do you think?

jreidinger: so different dialog. It is possible.

PR #401, opened 2019-01-30
Automatically select the driver packages (bsc#953522)

lslezak:
## Summay

The driver packages should be automatically preselected to install in installed system (during installation it already works), see https://bugzilla.suse.com/show_bug.cgi?id=953522.

## Solution

- We need to set a specific solver mode and let the solver preselect the system packages.
- We need to only preselect the drivers from new repositories (to not offer them again if user deselects them).
- See https://github.com/yast/yast-packager/wiki/Selecting-the-Driver-Packages for more details

## Testing

- Added unit tests
- Tested in installed system
  - Added a new repository in the repository manager and started the package manager,  the new drivers were preselected
  - Started the repository manager directly from the package manager
- Tested in installation
  - After the repositories used during installation were marked as "known" in the installed system. (Implemented in https://github.com/yast/yast-installation/pull/774)

## References 

- https://github.com/yast/yast-packager/wiki/Selecting-the-Driver-Packages
- https://bugzilla.suse.com/show_bug.cgi?id=953522
- https://github.com/yast/yast-installation/pull/774

### DIFF
--- a/src/lib/y2packager/known_repositories.rb
+++ b/src/lib/y2packager/known_repositories.rb
@@ -37,9 +37,7 @@ module Y2Packager
 
       # accessible only for the root user, the repository URLs should not contain
       # any passwords but rather be safe than sorry
-      File.open(status_file, "w", 0o600) do |f|
-        f.write(repositories.to_yaml)
-      end
+      File.write(status_file, repositories.to_yaml, open_args: ["w", 0o600])
     end
 
     def update
--- a/src/lib/y2packager/known_repositories.rb
+++ b/src/lib/y2packager/known_repositories.rb
@@ -1,0 +1,98 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2019 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "yaml"
+require "yast"
+
+module Y2Packager
+  # Track the known repositories from which the system packages (drivers)
+  # have been installed (or suggested to the user).
+  # @see https://github.com/yast/yast-packager/wiki/Selecting-the-Driver-Packages
+  class KnownRepositories
+    include Yast::Logger
+
+    STATUS_FILE = "/var/lib/YaST2/system_packages_repos.yaml".freeze
+
+    # Constructor
+    def initialize
+      Yast.import "Pkg"
+      Yast.import "Installation"
+    end
+
+    def repositories
+      @repositories ||= read_repositories
+    end
+
+    def write
+      log.info("Writing known repositories #{repositories.inspect} to #{status_file}")
+
+      # accessible only for the root user, the repository URLs should not contain
+      # any passwords but rather be safe than sorry
+      File.open(status_file, "w", 0o600) do |f|
+        f.write(repositories.to_yaml)
+      end
+    end
+
+    def update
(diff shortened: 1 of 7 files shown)


src/lib/y2packager/known_repositories.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    def write
+      log.info("Writing known repositories #{repositories.inspect} to #{status_file}")
+
+      # accessible only for the root user, the repository URLs should not contain
+      # any passwords but rather be safe than sorry
+      File.open(status_file, "w", 0o600) do |f|
+        f.write(repositories.to_yaml)
+      end
```
jreidinger: you can write it also as `File.write(status_file, repositories.to_yaml, open_args: ["w", 0o600])`

src/lib/y2packager/known_repositories.rb
```diff
@@ -0,0 +1,98 @@
...
+    #
+    # Return new (unknown) repositories
+    #
+    # @return [Array<String>] List of new repositories (URLs)
+    #
+    def new_repositories
+      log.info "current repositories: #{current_repositories.inspect}"
+      log.info "known repositories: #{repositories.inspect}"
+
+      new_repos = current_repositories - repositories
```
jreidinger: how expensive is calling new_repositories? you call it here twice and it calls Yast::Pkg.SourceGetCurrent(true) which I am not sure how expensive is.

src/lib/y2packager/system_packages.rb
```diff
@@ -0,0 +1,92 @@
...
+    #
+    # Create repository ID to URL mapping
+    #
+    # @return [Array<Integer>] List of repository IDs
+    #
+    def repo_ids(urls)
+      repo_ids = Yast::Pkg.SourceGetCurrent(true)
+      repo_ids.each_with_object([]) do |i, list|
+        list << i if urls.include?(Yast::Pkg.SourceGeneralData(i)["url"])
+      end
```
jreidinger: why `each_with_object`? it looks more like select for me. ` repo_ids.select { |i| urls.include?(Yast::Pkg.SourceGeneralData(i)["url"]) `

src/lib/y2packager/system_packages.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def find_packages
+      if repositories.empty?
+        log.info "No new repository found, not searching system packages"
+        return []
+      end
+
+      original_solver_flags = Yast::Pkg.GetSolverFlags
+
+      # solver flags for selecting minimal recommended packages (e.g. drivers)
```
jreidinger: please add there also bug report which contain additional sources to that recommended packages as it contain some additional info.

src/lib/y2packager/known_repositories.rb
lslezak: Ah, nice trick!

src/lib/y2packager/known_repositories.rb
lslezak: It just iterates over the already loaded repositories (`std::vector<>`). It's really cheap, it does not read/parse anything.

src/lib/y2packager/system_packages.rb
lslezak: Ah, I think I originally used some different structure here....

src/lib/y2packager/system_packages.rb
```diff
@@ -51,9 +51,7 @@ def select
     #
     def repo_ids(urls)
       repo_ids = Yast::Pkg.SourceGetCurrent(true)
-      repo_ids.each_with_object([]) do |i, list|
-        list << i if urls.include?(Yast::Pkg.SourceGeneralData(i)["url"])
-      end
+      repo_ids.select { |i| urls.include?(Yast::Pkg.SourceGeneralData(i)["url"]) }
```
jreidinger: when I see this just small note. In general I do not like much when method and local variable has identical name as in some cases it can be quite confusing (not mentioning clobbering). so what e.g. renaming local variable to `all_ids` ?

lslezak: oh, good point, I overlooked that...

PR #403, opened 2019-02-13
Explain selecting missing patterns (bsc#1117535)

mvidner:
- https://trello.com/c/QQJ9b8vB
- https://bugzilla.suse.com/show_bug.cgi?id=1117535

- only one error pop-up instead of one per each missing pattern
- suggest how to recover

TODO:
- ~~version+changelog~~
- ~~test the unhandled action~~

Bonus: code_coverage++ by dead_code--

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -2222,89 +2222,6 @@ module Yast
       ret
     end
 
-    # see bug 302398
-    def SelectKernelPackages
-      provides = Pkg.PkgQueryProvides("kernel")
-      # e.g.: [["kernel-bigsmp", `CAND, `NONE], ["kernel-default", `CAND, `CAND],
-      # ["kernel-default", `BOTH, `INST]]
-      Builtins.y2milestone("provides: %1", provides)
-
-      # these kernels would be installed
-      kernels = Builtins.filter(provides) do |l|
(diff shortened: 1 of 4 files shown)


src/modules/Packages.rb
```diff
@@ -2558,17 +2489,23 @@ def patterns_to_install
...
-      end
+    # @return [void]
+    def report_missing_patterns(pattern_names)
+      return if pattern_names.empty?
+
+      # "p1, p2, p3, p4, p5,\n" \
+      # "p6, p7, p8"
+      names_s = pattern_names.each_slice(5).map { |n5| n5.join ", " }.join ",\n"
+      # %s is a list of pattern names
+      Report.Error(_(
```
mvidner: The improved pop-up message is here.

src/modules/Packages.rb
```diff
@@ -2558,17 +2489,23 @@ def patterns_to_install
...
+        "Failed to select default product patterns:\n" \
+        "%s\n" \
+        "Patterns have not been found.\n" \
+        "\n" \
+        "This can be probably be fixed by adding\n" \
```
okurz: Sorry that I caught this only now but there is a duplication of "be" here :(

I hope my fix in https://github.com/yast/yast-packager/pull/405 is acceptable

PR #408, opened 2019-03-07
Changing repo URL has been ignored

schubi2:


src/modules/SourceDialogs.rb
```diff
@@ -2540,7 +2540,11 @@ def EditDisplayInt(repository)
       Builtins.y2milestone("Ret: %1", ret)
       UI.CloseDialog
 
-      ret == :ok ? GetURL() : ""
+      if ret == :ok || ret == :next
```
dgdavid: Maybe you can change this by

```ruby
[:ok, :next].include?(ret) ? GetURL() : ""
```
to make Rubocop happy

dgdavid: I mean, to use less lines and keep the class below the 2000 lines.

dgdavid: or even

```ruby
return GetURL() if ret == :ok || ret == :next

""
```

:see_no_evil:

schubi2: I like the first one :-)

PR #416, opened 2019-03-26
Fix: Checking if the proposal has been changed (right sorting).

schubi2:


lslezak:
Please add some more logging, see my commit https://github.com/yast/yast-packager/commit/261d61b81d314070c01c596f8ec5f5d75445788d that could help with debugging to find the change if there is any.

src/modules/Packages.rb
```diff
@@ -2420,6 +2353,62 @@ def remote_x11_packages
...
+      log_proposal_diff if changed
+      changed
+    end
+
+    #
+    # Log changed proposal
+    #
+    def log_proposal_diff
+      current = current_proposal
+      current.each_key do |kind|
```
lslezak: Maybe a simplification:

```ruby
current.each do |kind, list|
  log_array_diff(kind, list, @cached_proposal[kind])
end

File: src/modules/Packages.rb [hunk 2]
```diff
@@ -2381,7 +2381,32 @@ def current_proposal
...
-      current_proposal != @cached_proposal
+      changed = current_proposal != @cached_proposal
+      log_proposal_diff if changed
+      changed
+    end
+
+    #
+    # Log changed proposal
+    #
+    def log_proposal_diff
```
dgdavid: NP: This two methods looks a bit cryptic to me.

I would unify them in a single (and more specific) one, since `log_array_diff` is is only used here, what do you think?

src/modules/Packages.rb
```diff
@@ -2420,6 +2353,62 @@ def remote_x11_packages
 
   private
 
+    # Current package, pattern, product, patch and language selection.
+    #
+    # @return [Hash] selected packages, patterns, products,...
+    def current_proposal
+      selected_packages = Pkg.GetPackages(:selected, false)
+      selected_patterns = Pkg.ResolvableProperties("", :pattern, "").map do |p|
```
dgdavid: NP: I am not sure if it is a good idea, but what about having a method like 

```ruby
def fetch_selected(category)
  items = Pkg.ResolvableProperties("", category, "").select { |i| i["status"] == :selected }
  items.map { |i| i["name"] }.sort
end
```

which allows, 

```ruby
def current_proposal
  {
    "packages"  => Pkg.GetPackages(:selected, false).compact.sort,
    "patterns"  => fetch(:pattern),
    "products"  => fetch(:product),
    "patches"   => fetch(:patch),
    "languages" => [Pkg.GetPackageLocale].concat(Pkg.GetAdditionalLocales).compact.sort
  }
end
```

But I said, is just an idea, not sure if good or not.

src/modules/Packages.rb
```diff
@@ -2420,6 +2355,42 @@ def remote_x11_packages
 
   private
 
+    def fetch_selected(category)
+      items = Pkg.ResolvableProperties("", category, "").select { |i| i["status"] == :selected }
+      items.each_with_object([]) { |i, a| a << i["name"] }.sort
```
lslezak: Ah, after changing `map` to `select` in the line above you do not need `each_with_object` here:

```suggestion
      items.map{ |i| i["name"] }.sort
```

src/modules/Packages.rb
```diff
@@ -2420,6 +2355,42 @@ def remote_x11_packages
...
+
+    # Current package, pattern, product, patch and language selection.
+    #
+    # @return [Hash] selected packages, patterns, products,...
+    def current_proposal
+      { "packages"  => Pkg.GetPackages(:selected, false).compact.sort,
+        "patterns"  => fetch_selected(:pattern),
+        "products"  => fetch_selected(:product),
+        "patches"   => fetch_selected(:patch),
+        "languages" => [Pkg.GetPackageLocale].concat(Pkg.GetAdditionalLocales).compact.sort }
```
lslezak: Switch `[item].concat(array)` to `array.append(item)`:

```suggestion
        "languages" => Pkg.GetAdditionalLocales.append(Pkg.GetPackageLocale).compact.sort }
```

src/modules/Packages.rb
dgdavid: NP: @lslezak , yes, but using your `.each_with_object` suggestion the item names could be collected in a single iteration (without `nil`s) and then just sort them.

```ruby
result = []

Pkg.ResolvableProperties("", category, "").each_with_object(result) do |item|
  result << item["name"] if item["status"] == :selected
end

result.sort
```

src/modules/Packages.rb
schubi2: If you are calling current_proposal with this change multiple times you will get 
"languages"=>["en_US", "en_US"]} . instead of "languages"=>["en_US"]

src/modules/Packages.rb
```diff
@@ -2420,6 +2355,42 @@ def remote_x11_packages
...
+    def fetch_selected(category)
+      items = Pkg.ResolvableProperties("", category, "").select { |i| i["status"] == :selected }
+      items.each_with_object([]) { |i, a| a << i["name"] }.sort
+    end
+
+    # Current package, pattern, product, patch and language selection.
+    #
+    # @return [Hash] selected packages, patterns, products,...
+    def current_proposal
+      { "packages"  => Pkg.GetPackages(:selected, false).compact.sort,
```
dgdavid: Just a question: it is `compact` really needed here? I was taking a look to the `GetPackages` code and I am not sure that it is returning `nil` values.

src/modules/Packages.rb
schubi2: Thanks to the testcases :-)

src/modules/Packages.rb
schubi2: Not really. I have removed it.

src/modules/Packages.rb
lslezak: So do we need appending `uniq`?

lslezak: Ah, duplicates do not matter, if they are always the same...

PR #422, opened 2019-04-03
Reset solver state before starting the next solver run

schubi2:


package/yast2-packager.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Apr  3 10:38:03 CEST 2019 - schubi@suse.de
+
+- Package proposal: Reset solver state before starting the next
+  solver run (bsc#1077399).
```
lslezak: NP: It's not fixing exactly that bug so I'd prefer a comment like 
```suggestion
  solver run (related to bsc#1077399).
```

PR #428, opened 2019-04-26
Display better product names (related to bsc#1133215)

lslezak:
...when upgrading from SLES12 + SUMA Proxy + SUMA Branch Server

- Display human readable product names (instead of internal IDs)
- 4.1.38

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -734,6 +734,7 @@ module Yast
 
       ret += status[:removed].map do |product|
         obsolete = Y2Packager::ProductUpgrade.will_be_obsoleted_by(product["name"])
+        log.info "Product #{product["name"].inspect} will be obsoleted by #{obsolete.inspect}"
         if obsolete.empty?
           transact_by = product["transact_by"]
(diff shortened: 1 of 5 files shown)


src/modules/Packages.rb
```diff
@@ -749,9 +750,18 @@ def product_update_summary(products)
...
           re = HTML.Colorize(msg, "red")
         else
+          selected_products = Y2Packager::Product.with_status(:selected)
+          # get the human readable product names from libzypp data
+          new_products = obsolete.map do |p|
+            new_product = selected_products.find { |selected_prod| selected_prod.name == p }
+            # fallback to the internal product name if not found (should never happen™,
+            # ProductUpgrade.will_be_obsoleted_by should return only the selected products)
+            new_product ? new_product.display_name : p
+          end
```
jreidinger: hmm, I know it is something for SP1, but I do not like much adding to method with 57 lines another 10 lines. Is possible to safely move it to own method?

jreidinger:
code wise it looks good, but that increase of big method is something that worry me a bit.

PR #431, opened 2019-05-03
External sources link, master

mvidner:
Merging #430

### DIFF
--- a/src/lib/y2packager/clients/inst_productsources.rb
+++ b/src/lib/y2packager/clients/inst_productsources.rb
@@ -445,7 +445,8 @@ module Yast
 
     def ReadControlFile
       # Prefer control.xml to /etc/YaST2/ProductFeatures (bsc#1132613)
-      ProductControl.Init
+      # (The latter file does not exist during the installation)
+      ProductControl.Init unless Stage.initial
       # Notably GetSection does not call ProductFeatures.Restore
       # which would overwrite the ProductControl data we've just read
       software_features = ProductFeatures.GetSection("software")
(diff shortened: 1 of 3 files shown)


src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -444,27 +444,23 @@ def InitializeSources
     end
 
     def ReadControlFile
+      # Prefer control.xml to /etc/YaST2/ProductFeatures (bsc#1132613)
+      ProductControl.Init
```
lslezak: This client is also called [during installation](https://github.com/yast/skelcd-control-openSUSE/blob/master/control/control.openSUSE.xml#L919). Is it safe to call [ProductControl.Init](https://github.com/yast/yast-yast2/blob/master/library/control/src/modules/ProductControl.rb#L1175) in that case?

mvidner: Uh, I don't know :-(

mvidner: I've checked openQA how #430 is doing. `productsources` is only enabled in openSUSE, so the oenQA group to check is https://openqa.opensuse.org/group_overview/50 "Last Builds for openSUSE Leap 15". It works fine.
The reinitialization could break add-ons, but they are only brought in later.

mvidner: As for logging, ProductControl.Init does call `Builtins.y2milestone("Reading control file: #{@current_control_file}")`

PR #434, opened 2019-05-07
Memory optimizations (bsc#1132650)

lslezak:
- Avoid freezing the installer when running with 1GB RAM and using the online repositories.
- Do not use the `Pkg.ResolvableProperties("", :package, "")` call which requires huge amount of memory 
  - Replaced with `Pkg.GetPackages` and `Pkg.ResolvableProperties(<pkg>, :package, "")`
  - This skips the not selected packages and also queries the selected packages one-by-one, it avoids creating a huge list with all packages at once
- Tested manually, see the screencasts below.
- Added an unit test
- 4.1.41

## Screencasts

Tested in a VM with 1GB RAM with graphical installation using the NET ISO.

### The Original Code

After selecting a role the installer stopped responding and even the mouse cursor was lagging.

low_memory_orig

### With the Fix Applied

With the fix the installation after a while normally continues to the partitioner proposal.

low_memory_fix

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -821,11 +821,16 @@ module Yast
     # @param [Array<Symbol>] keep a list of symbols specifying type of objects to be kept selected
     def Reset(keep)
       restore = []
+      restore_packages = []
 
       # collect the currently selected resolvables
       keep.each do |type|
-        resolvables = Pkg.ResolvableProperties("", type, "")
+        # handle packages differently, there might be tens of thousands packages,
+        # the package list might be HUGE (bsc#1132650)
+        # but there are usually just few dozens of products, patterns, etc.
+        next if type == :package
 
+        resolvables = Pkg.ResolvableProperties("", type, "")
         resolvables.each do |resolvable|
           # only selected items but ignore the selections done by solver,
           # during restoration they would be changed to be selected by YaST and they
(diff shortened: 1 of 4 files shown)


src/modules/Packages.rb
```diff
@@ -836,10 +841,25 @@ def Reset(keep)
...
 
+      # special handling for packages, there might be tens of thousands packages
+      # so we need to do it more effectively
+      if keep.include?(:package)
+        # first query the selected packages, then get the details for each selected
+        # package separately to avoid processing a huge list at once
+        # (true = names only without version)
+        Pkg.GetPackages(:selected, true).each do |pkg|
+          Pkg.ResolvableProperties(pkg, :package, "").each do |resolvable|
+            next if resolvable["status"] != :selected || resolvable["transact_by"] == :solver
```
mvidner: I dislike the filtering logic duplication here. I have an idea to do all this in a different way, by constructing a different enumerator for `resolvables` above, but need to check if it works at all and is not even more complicated than this.

lslezak: I also noticed the duplication but as this is for SP1 I rather prefer a simple although not a perfect fix.

mvidner: OK, here's my experiment with a lazy enumerator: https://gist.github.com/mvidner/551d56d676e5700db6926c480276ee37

src/modules/Packages.rb
```diff
@@ -836,10 +841,25 @@ def Reset(keep)
...
+            restore_packages << pkg
```
mvidner: I think the fix can be made simpler if we drop `restore_packages` and say `restore << [pkg, :package]` here

lslezak: I deliberately avoided that, there might be thousands of packages, creating extra nested array (a pair) `[pkg, :package]` for each package will take additional memory. Having a simple `String` value instead of nested `Array` might make some memory difference.

(Though I do not have any real numbers. AFAIK the symbols are shared in Ruby so that should not be a big problem, but I'm not sure how Ruby `Array` is effective, esp. if it contains just two items.)

PR #436, opened 2019-05-14
Sle 15 sp1 bsc 1132622

schubi2:


src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -1496,7 +1488,10 @@ def CreateSource(url, pth, repo_name, alias_s)
         Builtins.y2milestone("Product Data: %1", prod)
 
         repo_id = CreateRepoId(url, pth)
-        Builtins.y2milestone("Addind repository with ID: %1", repo_id)
+        Builtins.y2milestone("Adding repository with ID: %1", repo_id)
+
+        # replace $releasever by the real version
+        url.gsub!("$releasever", Product.version)
```
lslezak: Use `Pkg.ExpandedUrl` also here.

src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -923,6 +911,10 @@ def PrintRepositoryDescription
         ""
       end
 
+      # replace $releasever by the real version
+      url_string = @list_of_repos[current_id]["url"] || ""
+      url_string.gsub!("$releasever", Product.version)
```
lslezak: Use `Pkg.ExpandedUrl` also here.

src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -377,23 +372,17 @@ def NormalizeURL(url_string)
     def IsAddOnAlreadySelected(s_url, s_path)
       ret = -1
 
-      s_url = NormalizeURL(s_url)
+      search_url = NormalizeURL((s_url || "") + (s_path || ""))
+      # replace $releasever by the real version
+      search_url.gsub!("$releasever", Product.version)
```
lslezak: Use [Pkg.ExpandedUrl](
https://github.com/yast/yast-pkg-bindings/blob/78f5ce710aa8973f50227150ed570467d0d71885/src/PkgFunctions.cc#L514) to get the full URL. It uses the libzypp functionality.

It's not that trivial as it looks at the first sight. There are also other variables like `$arch`, `$releasever_major`, `$releasever_minor` and others. There might be even custom defined values. See the `Variable substitution` section in `man zypper`.

src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -358,17 +359,11 @@ def NetworkRunning
...
-      if Builtins.regexpmatch(url_string, "%")
-        # unescape it
-        url_string = URL.UnEscapeString(url_string, URL.transform_map_filename)
-      end
+      # removing all slashes at the end of the url
+      url_string.sub!(/(\/)+$/, "")
 
-      url_string
+      # unescape it
+      URI.unescape(url_string)
```
lslezak: Just a theoretical question, what if the slash at the end is escaped? I think first we should unescape and then remove the trailing slashes...

src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -412,10 +401,9 @@ def InitializeSources
       Pkg.TargetInitialize(Installation.destdir)
       # the fastest way
       Pkg.SourceRestore
-
       if !Mode.installation
         # repos_already_used
-        Builtins.foreach(Pkg.SourceGetCurrent(true)) do |one_id|
+        Builtins.foreach(Pkg.SourceGetCurrent(false)) do |one_id|
```
lslezak: OK, so also check the disabled repositories...

schubi2: Yes, I think that should be OK.

test/lib/clients/inst_productsources_test.rb
```diff
@@ -110,4 +111,96 @@ def list_of_repos
...
+        }
+      ]
+    end
+
+    before do
+      allow(Yast::AddOnProduct).to receive(:add_on_products).and_return(add_on_products)
+    end
+
+    context "is already added" do
+      it "returns source id" do
```
lslezak: NP: Um, what the the test does is actually checking that the empty or default product path works correctly. So maybe something like this would be better:

```suggestion
      it "handles the default and empty product directory correctly" do
```

PR #441, opened 2019-05-22
Allow to select the license language when running in text mode

dgdavid:
- https://bugzilla.suse.com/show_bug.cgi?id=1135901 (was P1)
- https://trello.com/c/pfv5PzvU

Problem: Only in the text mode installation, we cannot switch the language of a product license. Both for the main product and add-on products. [bsc#1135901](https://bugzilla.suse.com/show_bug.cgi?id=1135901)

Reason: That's on purpose, since on the Linux console we're not able to display the characters of most languages, [bsc#1094793](https://bugzilla.suse.com/show_bug.cgi?id=1094793). But wait, recently we have switched to starting fbiterm unconditionally, so we are in fact able to display all languages that currently have license translations. So let's enable the language switch.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# 2 space indentation
+[*.rb]
+indent_style = space
+indent_size = 2
--- a/src/lib/language_tag.rb
+++ b/src/lib/language_tag.rb
@@ -81,9 +81,10 @@ class LanguageTag
     # FIXME: or self, find out what makes more sense
   end
 
-  # @return [String,nil]
+  # @return [String,nil] a translated name, eg. "Tschechisch" for "cs" or "cs_CZ"
   def name(lang_map_cache: nil)
-    lang_map_cache ||= Yast::Language.GetLanguagesMap(false)
+    force_retranslation = false
+    lang_map_cache ||= Yast::Language.GetLanguagesMap(force_retranslation)
     attrs = lang_map_cache[@tag]
     if attrs.nil?
       # we're en, find en_US
--- a/src/lib/language_tag.rb
+++ b/src/lib/language_tag.rb
@@ -1,0 +1,99 @@
+# ------------------------------------------------------------------------------
+# Copyright (c) 2019 SUSE LLC, All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of version 2 of the GNU General Public License as published by the
+# Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+# ------------------------------------------------------------------------------
+
+require "yast"
+
+# {::Comparable} enforces a total ordering, contrary to its
+# documentation, WTF.
+module PartiallyComparable
+  def <(other)
+    cmp = self.<=>(other)
+    return nil if cmp.nil?
+    cmp < 0
+  end
+
+  def >(other)
+    cmp = self.<=>(other)
+    return nil if cmp.nil?
+    cmp > 0
+  end
+
+  def <=(other)
+    cmp = self.<=>(other)
+    return nil if cmp.nil?
+    cmp <= 0
+  end
+
+  def >=(other)
+    cmp = self.<=>(other)
+    return nil if cmp.nil?
+    cmp >= 0
+  end
+
+  def ==(other)
+    return true if equal?(other) # object identity
+    cmp = self.<=>(other)
+    return nil if cmp.nil?
+    cmp == 0
+  end
+end
+
+# Language tags like "cs" "cs_CZ" "cs_CZ.UTF-8".
+#
+# FIXME: improve the simplistic string comparisons
+class LanguageTag
+  include Yast::Logger
+
+  # @param s [String]
+  def initialize(s)
+    @tag = s
+  end
+
+  def to_s
+    @tag
+  end
+
+  include PartiallyComparable
+
+  # Like with classes (where Special < General) "en_US" < "en"
+  # Mnemonics: number of speakers
+  def <=>(other)
+    return 0 if to_s == other.to_s
+    return -1 if to_s.start_with?(other.to_s)
+    return 1 if other.to_s.start_with?(to_s)
+    nil
+  end
+
+  # A more general tag: "en_US" -> "en" (-> nil)
+  # @return [LanguageTag,nil]
+  def generalize
+    self.class.new(@tag.split("_").first) if @tag.include? "_"
+    # else nil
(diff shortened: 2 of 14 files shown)


mvidner:
See also another check for TextMode: https://github.com/yast/yast-packager/blob/71d43cfad3e50302467df792f1748d492b943576/src/modules/ProductLicense.rb#L681

src/lib/y2packager/widgets/product_license_translations.rb
```diff
@@ -126,6 +126,15 @@ def default_language
...
         DEFAULT_FALLBACK_LANGUAGE
       end
+
+      # Whether it is needed to force the default language
+      #
+      # @return [Boolean] true if running but not in a fbiterm; false otherwise
+      def force_language?
+        return false unless Yast::UI.TextMode
+
+        Yast::Builtins.getenv("TERM") == "iterm"
```
mvidner: The other way around!

src/lib/y2packager/widgets/product_license_translations.rb
```diff
@@ -126,6 +126,15 @@ def default_language
         return candidate_lang if translated
         DEFAULT_FALLBACK_LANGUAGE
       end
+
+      # Whether it is needed to force the default language
+      #
+      # @return [Boolean] true if running but not in a fbiterm; false otherwise
```
mvidner: if running what

src/lib/y2packager/widgets/product_license_translations.rb
```diff
@@ -126,6 +126,15 @@ def default_language
...
+      def force_language?
+        return false unless Yast::UI.TextMode
```
lslezak: You can simplify it to a single line and even avoid using the obsolete `Yast::Builtins` call:

```ruby
Yast::UI.TextMode && ENV["TERM"] != "iterm"
```

test/lib/widgets/product_license_translations_test.rb
```diff
@@ -45,6 +45,7 @@
 
       before do
         allow(Yast::UI).to receive(:TextMode).and_return(true)
+        allow(Yast::Builtins).to receive(:getenv).with("TERM").and_return("iterm")
```
mvidner: we should split this context instead of modifying it:
- when running on textmode, with iterm
- when running on textmode, without iterm

dgdavid:
@mvidner, @lslezak Now the PR is ready to review. Thanks!

mvidner:
There are more checks for TextMode in ProductLicense: https://github.com/yast/yast-packager/blob/71d43cfad3e50302467df792f1748d492b943576/src/modules/ProductLicense.rb#L681

and https://github.com/yast/yast-packager/blob/71d43cfad3e50302467df792f1748d492b943576/src/modules/ProductLicense.rb#L1713

mvidner:
My WIP is https://github.com/yast/yast-packager/commit/c067f5ddf17c68d9369fea39734812f1d4f43292 It fails the old unit tests, and changes only the main product, but its ideas are clear enough to be transferable to the add-on product case in ProductLicense.rb

mvidner:
TODO: things to test:
- license available
- language displayable
- lang_code_match("zh", "zh_CN"): is it symmetric? "pt_BR" vs "pt-br"
- "" is English (WTF)

kobliha:
@mvidner `""` is `"en_US"` as that one has been chosen as the default language.

test/lib/widgets/product_license_translations_test.rb
```diff
@@ -17,85 +17,76 @@
 require "y2packager/widgets/product_license_translations"
 require "y2packager/product"
 
+RSpec::Matchers.define :array_not_including do |x|
+  match do |actual|
+    return false unless actual.is_a?(Array)
+    return false if actual.include?(x)
```
lslezak: Just a simple `!actual.include?(x)` instead of the 2 lines?

.editorconfig
```diff
@@ -0,0 +1,9 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# 2 space indentation
+[*.rb]
+indent_style = space
+indent_size = 2
```
lslezak: What about adding also `trim_trailing_whitespace = true`?

- I'm not sure if we want to add it as a part of a maintenance update, maybe adding only to `master` would be better.
- It would make sense to have it in *all* repositories, but then we should agree on the exact content first (discuss at yast-devel@ ?)

src/lib/language_tag.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  # A more general tag: "en_US" -> "en" (-> nil)
+  # @return [LanguageTag,nil]
+  def generalize
+    self.class.new(@tag.split("_").first) if @tag.include? "_"
+    # else nil
+    # FIXME: or self, find out what makes more sense
+  end
+
+  # @return [String,nil]
```
lslezak: Some more comment (what the method does/returns) would be nice as it is not obvious what `attrs[4]` stands for.

test/lib/widgets/product_license_translations_test.rb
```diff
@@ -17,85 +17,74 @@
 require "y2packager/widgets/product_license_translations"
 require "y2packager/product"
 
+RSpec::Matchers.define :array_not_including do |x|
```
lslezak: Can this new rspec matcher leak to the other tests executed *after* this one? (When using `rake test:unit`.) Would it be then better to share it explicitly in the `test_helper.rb`?

test/lib/widgets/product_license_translations_test.rb
```diff
@@ -17,85 +17,74 @@
...
-            .with([language], language)
-          widget.contents
-        end
-
-        it "shows the product license in the default language" do
-          expect(Y2Packager::Widgets::ProductLicenseContent).to receive(:new)
-            .with(product, language)
-          widget.contents
-        end
+      it "shows the product license in the default language (AmE)" do
```
lslezak: For me `en_US` would more readable here...
```suggestion
      it "shows the product license in the default language (en_US)" do
```

.editorconfig
mvidner: OK, I'll bring it up on the mailing list

test/lib/widgets/product_license_translations_test.rb
dgdavid: Yes, I agree sharing it explicitly sharing it in the test helper. Just I forget to do it because I wanted to check first if it is really necessary o there is another approach.

src/lib/language_tag.rb
```diff
@@ -0,0 +1,99 @@
...
+    return -1 if to_s.start_with?(other.to_s)
+    return 1 if other.to_s.start_with?(to_s)
+    nil
+  end
+
+  # A more general tag: "en_US" -> "en" (-> nil)
+  # @return [LanguageTag,nil]
+  def generalize
+    self.class.new(@tag.split("_").first) if @tag.include? "_"
+    # else nil
```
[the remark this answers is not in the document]
lslezak: Um, I'd actually put some comment that `nil` might be returned above the code line. If you just quickly scan the line you might miss that `if` at the end and think that an object is always returned.

PR #442, opened 2019-05-27
Added new Resolvable class, use it in Packages.rb (bsc#1136051)

lslezak:
- This is continuation of the https://github.com/yast/yast-pkg-bindings/pull/114 change
- Added new `Resolvable` class
- Use the new class to fix [bug#1136051](https://bugzilla.suse.com/show_bug.cgi?id=1136051)
- Added unit test


src/lib/y2packager/resolvable.rb
```diff
@@ -0,0 +1,133 @@
...
+    # @param hash [Hash<Symbol,Object>] A pkg-bindings resolvable hash.
+    def initialize(hash)
+      from_hash(hash)
+    end
+
+    #
+    # Dynamically load the missing attributes from libzypp.
+    #
+    # @param method [Symbol] the method called
+    # @param _args no used so far
```
jreidinger: maybe raise NOMETHOD error imediatelly if args is passed? It basically makes `a.version` identical to `a.version(force: true, blabla: "Yes")`

src/lib/y2packager/resolvable.rb
```diff
@@ -0,0 +1,135 @@
...
+    # @return [Object] The read value.
+    def load_attribute(attr)
+      attrs = Hash[(UNIQUE_ATTRIBUTES.map { |a| [a, instance_variable_get("@#{a}")] })]
+      resolvables = Yast::Pkg.Resolvables(attrs, [attr])
+
+      # Finding more than one result is suspicious, log a warning
+      log.warn("Found several resolvables: #{resolvables.inspect}") if resolvables.size > 1
+
+      resolvable = resolvables.first
+      raise NoMethodError unless resolvable && resolvable.key?(attr.to_s)
```
jreidinger: this is a bit problematic, because exception contains no details.

I suggest to return here instead nil and above in `method_missing` call `super` if nil is returned. So it raise proper one. Just compare:

```
irb(main):001:0> def a
irb(main):002:1> raise NoMethodError
irb(main):003:1> end
irb(main):006:0> Object.a
        3: from /usr/bin/irb:11:in `<main>'
        2: from (irb):6
        1: from (irb):2:in `a'
NoMethodError (NoMethodError)
irb(main):007:0> Object.b
        2: from /usr/bin/irb:11:in `<main>'
        1: from (irb):7
NoMethodError (undefined method `b' for Object:Class)
```

PR #450, opened 2019-06-18
Sle 15 sp1 bsc 1136708

schubi2:
- https://bugzilla.suse.com/show_bug.cgi?id=1136708
- https://trello.com/c/hirlFUmw

While showing the slideshow different YAST modules will be called:
Calling YaST client inst_prepareprogress
Calling YaST client wrapper_slideshow_callbacks
Calling YaST client inst_prepdisk
Calling YaST client inst_instsys_cleanup
Calling YaST client inst_kickoff
Calling YaST client inst_bootloader
Calling YaST client inst_rpmcopy
...
..
.
The problem is that not each of these modules handles UI callbacks. Switching to another tab like the release notes tab will not be supported there.
With this fix the release tab will be shown only while RPM packages will be installed. So it will be shown 3-4 seconds after starting the installation.

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -2,7 +2,7 @@
 Mon Jun 17 17:15:29 CEST 2019 - schubi@suse.de
 
 - Package installation: Rebuild slide show dialog and enable
-  realease tab (bsc#1136708).
+  realease notes tab (bsc#1136708).
 - 4.1.44
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


mvidner:
> The problem is that not each of these modules handles UI callbacks

You mean, they are not running a UI event loop, in particular not the one event loop that responds to the Details and ReleaseNotes tabs, right?

schubi2:
> > The problem is that not each of these modules handles UI callbacks
> 
> You mean, they are not running a UI event loop, in particular not the one event loop that responds to the Details and ReleaseNotes tabs, right?

yes, your explanation is better than mine :+1:

PR #452, opened 2019-06-18
Allow to use "refuse" as a cancel action

dgdavid:
## Problem

The `ProductLicense.HandleLicenseDialogRet` do not have a cancel action to indicate that the user wants to refuse an EULA.

Related to https://bugzilla.suse.com/show_bug.cgi?id=1114018.

## Solution

To give support for a "refuse" cancel action, required by https://github.com/yast/yast-registration/pull/439

## Tests

* Added and updated some unit tests.
* Also tested manually via driver update.

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -1100,6 +1100,8 @@ module Yast
           log.info "License has been declined."
 
           case action
+          when "refuse"
+            ret = :refused
           when "abort"
             ret = :abort
           when "halt"
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jun 18 08:59:39 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Better handling of license agreement dialog, allowing to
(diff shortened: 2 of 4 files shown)


src/modules/ProductLicense.rb
```diff
@@ -1100,6 +1100,8 @@ def HandleLicenseDialogRet(licenses, base_product, action)
           log.info "License has been declined."
 
           case action
+          when "refuse"
+            ret = :refuse
```
lslezak: Probably "refuse**d**" is a bit better...
```suggestion
            ret = :refused
```

dgdavid: Ups, sorry.

I was updated it due to your comments https://github.com/yast/yast-registration/pull/439.

I also changed the "cancel_action", not only the "ret" value. What do you think?

lslezak: I'd keep the cancel action"refuse" (means what the code should do when canceled, it's an action), and have the result "refused" (what the user did, it's the result).

It would nice to also have the "abort"/"aborted" pair but as this would break the backward compatibility we should not change that.

dgdavid:
I solved the conflict and changed from `3.3.0` to `3.3.1` version.

PR #459, opened 2019-07-11
Improve handlig Packages DVD/Offline installation medium

lslezak:
- Improved product selection for multi-repositorie media
  - Display more product details
  - Handle the product dependencies
- Travis (unit tests) fails because of a missing `ruby-solv` package, works fine locally

#### Graphical (dark SLE installation theme)

offline_addons

#### Text Mode

new_addon_ncurses


package/yast2-packager.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Jul 11 07:17:45 UTC 2019 - Ladislav Slezák <lslezak@suse.cz>
+
+- Improved product selection for multi-repositorie media
```
jreidinger: typo multi-repositorie

package/yast2-packager.changes
```diff
@@ -1,3 +1,11 @@
...
+  - Display more product details
+  - Handle the product dependencies
```
jreidinger: jsc, bsc or fate missing

src/include/packager/repositories_include.rb
```diff
@@ -340,9 +341,17 @@ def autorefresh_for?(url)
...
-    def scan_products(expanded_url, original_url)
-      new_repos = Pkg.RepositoryScan(expanded_url)
-      found_products = new_repos.map { |(name, dir)| Y2Packager::ProductLocation.new(name, dir) }
+    def scan_products(_expanded_url, original_url)
+      # use the selected base product during installation,
+      # in installed system use the installed base product
+      base_product = if Stage.initial
+        Y2Packager::Product.selected_base.name
+      else
+        Y2Packager::Product.installed_base_product.name
```
jreidinger: when I see this API, it looks inconsistent, why it is selected_base and installed_base_product? both should be base_product or both only base.

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -37,6 +43,9 @@ def initialize(products)
         textdomain "packager"
 
         @products = products
+        # do not offer base products, they would conflict with the already selected base product,
+        # allow a hidden way to force displaying them in some special cases
+        @products.reject!(&:base) unless ENV["Y2_DISPLAY_BASE_PRODUCTS"] == "1"
```
jreidinger: do not forget to document that new env variable.

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -63,13 +72,21 @@ def abort_handler
...
+      # Handle changing the current item or chaging the selection
+      def addon_repos_handler
+        current_item = Yast::UI.QueryWidget(Id(:addon_repos), :CurrentItem)
+        current_product = products.find { |p| p.dir == current_item }
+
+        return unless current_product
+
+        refresh_details(current_product)
+
+        select_dependant_products
```
jreidinger: well, spelling of dependant catch my eyes, so I check it and looks like according to british english it should be used only as noun. In american it does not matter https://grammarist.com/spelling/dependant-dependent/

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -107,9 +124,10 @@ def dialog_content
           # TRANSLATORS: Product selection label (above a multi-selection box)
           Left(Heading(_("Available Extensions and Modules"))),
 
-          VWeight(75, MinHeight(12,
+          VWeight(60, MinHeight(8,
```
jreidinger: can you also add screenshot of ncurses version? as screen looks a bit busy for me.

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -119,14 +137,51 @@ def dialog_content
...
+        # select the dependent products
+        new_selection = current_selection
+
+        # the selection has not changed, nothing to do
+        return if new_selection == selected_products
+
+        # add the dependant items to the selected list
+        selected_items = Yast::UI.QueryWidget(Id(:addon_repos), :SelectedItems)
+        new_items = new_selection - selected_products
+        new_items.each do |p|
```
jreidinger: NP: you can write it as 
```
selected_items = new_items.each_with_object(Yast::UI.QueryWidget(Id(:addon_repos), :SelectedItems)) do |p, res|
  res.concat(p.depends_on)
end
```

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -147,17 +202,35 @@ def continue_msg
...
+      end
+
+      def product_description(product)
+        erb_file = File.join(__dir__, "product_summary.erb")
+        log.info "Loading ERB template #{erb_file}"
+        erb = ERB.new(File.read(erb_file))
+
+        # compute the dependant products
+        dependencies = []
+        if product.depends_on && !product.depends_on.empty?
```
jreidinger: second part of condition is not needed, as each for empty array is ignored.

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -147,17 +202,35 @@ def continue_msg
...
+          product.depends_on.each do |p|
```
jreidinger: so with new rubocop merged (not yet), you can use `product.depends_on&.each` and avoid condition completelly.

src/lib/y2packager/product_finder.rb
```diff
@@ -0,0 +1,139 @@
...
+    #
+    # @param base [String] The name of the base product used for evaluating the
+    #  dependencies.
+    #
+    # @return [Array<Hash>] The list of found products
+    #
+    def products(base = nil)
+      marked_base_products = base_product_tags
+      # evaluate all products
+      pool.whatprovides(pool.str2id("product()")).each_with_object([]) do |s, list|
```
jreidinger: this block is a bit too big. I suggest to simply have something like
```
pool.whatprovides(pool.str2id("product()")).each_with_object([]) do |s, list|
  list.concat(create_products(s)
end
```

btw that "product()" looks like string.

src/lib/y2packager/product_finder.rb
```diff
@@ -0,0 +1,139 @@
...
+        if base
+          # select the base product
+          base_product = pool.select("product(#{base})", Solv::Selection::SELECTION_PROVIDES)
+          jobs += base_product.jobs(Solv::Job::SOLVER_INSTALL)
+        end
+
+        # run the solver to evaluate all dependencies
+        problems = solver.solve(jobs)
+
+        # in case of problems, ignore the dependencies
```
jreidinger: why ignore in case of problems?

src/lib/y2packager/product_finder.rb
```diff
@@ -0,0 +1,139 @@
...
+        if problems.empty?
+          # find all repositories which have a product selected to install
+          solver.transaction.newsolvables.each do |n|
+            next if n == s
```
jreidinger: single letters for block is only good for small scope. Here it is confusing.

src/lib/y2packager/product_finder.rb
```diff
@@ -0,0 +1,139 @@
...
+          description:     s.lookup_str(Solv::SOLVABLE_DESCRIPTION),
+          depends_on:      problems.empty? ? required : nil,
+          order:           display_order(s)
+        }
+
+        # in theory a release package might provide several products,
+        # create an item for each of them
+        s.lookup_deparray(Solv::SOLVABLE_PROVIDES).each do |p|
+          next unless p.str =~ /\Aproduct\(\)\s*=\s*(\S+)/
+          product_name = Regexp.last_match[1]
```
jreidinger: another option can be
```
product_name = p.str[/\Aproduct\(\)\s*=\s*(\S+)/, 1]
next unless product_name
```

src/lib/y2packager/product_finder.rb
```diff
@@ -0,0 +1,139 @@
...
+  private
+
+    #
+    # Return the list of marked base products. A base product is defined
+    # by the "system-installation() = <product>" provides.
+    #
+    # @return [Array<String>] The base products
+    #
+    def base_product_tags
+      install_provides = pool.whatprovides(pool.str2id("system-installation()"))
```
jreidinger: another constant

src/lib/y2packager/product_finder.rb
```diff
@@ -0,0 +1,139 @@
...
+    # @return [Integer,nil] The display order value or nil if not defined
+    #
+    def display_order(solvable)
+      # all solvable provides
+      provides = solvable.lookup_deparray(Solv::SOLVABLE_PROVIDES)
+
+      order = nil
+      provides.each do |p|
+        next unless p.str =~ /\Adisplayorder\(\s*([0-9]+)\s*\)\z/
+        order = Regexp.last_match[1].to_i if Regexp.last_match
```
jreidinger: is that `if Regexp.last_match` needed?

src/lib/y2packager/product_location.rb
```diff
@@ -10,22 +10,95 @@
...
 
+    # @return [String] Product name (the zypp resolvable)
+    attr_reader :product_name
+    # @return [String] The product package (*-release RPM usually)
+    attr_reader :product_package
+    # @return [String] Product summary
+    attr_reader :summary
+    # @return [String] Product description
+    attr_reader :description
+    # @return [Integer] Display order
```
jreidinger: from code it looks like it can be also nil. What happens in that case?

src/lib/y2packager/product_location.rb
```diff
@@ -10,22 +10,95 @@
...
+      log.info "Scanning #{Yast::URL.HidePassword(url)} for products..."
+
+      downloader = Y2Packager::RepomdDownloader.new(url)
+      pool = Y2Packager::SolvablePool.new
+
+      repomd_files = downloader.download
+      return [] if repomd_files.empty?
+
+      repomd_files.each do |repoindex|
+        pool.add_rpmmd_repo(repoindex, "/" + repoindex.split("/")[-3])
```
jreidinger: this -3 looks very magic, can you add comment explaining it?

src/lib/y2packager/product_location.rb
```diff
@@ -10,22 +10,95 @@
...
+      end
+    end
+
     # Constructor
     #
     # @param name [String] Product name
     # @param dir [String] Location (path starting at the media root)
-    def initialize(name, dir)
+    def initialize(name, dir, product_name: nil, summary: nil, product_package: nil,
+      order: nil, description: nil, base: nil, depends_on: nil)
```
jreidinger: so basically almost any attribute can be nil. Does not make sense to have some kind of defaults here?

jreidinger: What worries me is that it means we need to check for nil almost everywhere.

src/lib/y2packager/repomd_downloader.rb
```diff
@@ -0,0 +1,66 @@
...
+require "yast"
+
+Yast.import "Pkg"
+
+module Y2Packager
+  # This class downloads the primary.xml metadata file from an rpm-md repository.
+  # For multi-product repositories it downloads the files from all product
+  # subdirectories.
+  class RepomdDownloader
+    attr_reader :url, :product_repos
```
jreidinger: what about having product_repos lazy loaded?
like
```
def product_repos
  @product_repos if @product_repos
  
  download
  @product_repos
end
```
and having download private method?

src/lib/y2packager/repomd_downloader.rb
```diff
@@ -0,0 +1,66 @@
...
+      @url = url
+    end
+
+    #
+    # Download the primary.xml file(s) from the repository.
+    #
+    # @return [Array<String>] List of paths pointing to the downloaded primary.xml files,
+    #   returns an empty list if the URL or the repository is not valid.
+    #
+    def download
```
jreidinger: name is a bit confusing, what about something like `primary_xmls` as what you want is actually that files. Also that side-effect for product_repos is not nice, as I mention above.

jreidinger: and maybe caching of that primary_xmls? and also please document who is reponsible for deleting that files.

src/lib/y2packager/solvable_pool.rb
```diff
@@ -0,0 +1,40 @@
...
+    #
+    # Load repository metadata to the pool.
+    #
+    # @param primary_xml [String] Path to the primary.xml.gz file
+    # @param name [String] Name of the repository
+    def add_rpmmd_repo(primary_xml, name)
+      repo = pool.add_repo(name)
+      gz = open(primary_xml)
+      fd = Solv.xfopen_fd(primary_xml, gz.fileno)
+      gz.close
```
jreidinger: close in ensuse if that xfopen failed?

src/modules/Packages.rb
```diff
@@ -2741,6 +2749,26 @@ def new_product_names(obsolete)
...
+    META_DATA_FILES = [
+      # repo-md repository
+      "repodata/repomd.xml",
+      # SUSE-tags repository
+      "content"
+    ].freeze
+
+    # repository metadata present at the url?
+    # @return [Boolean] true if a repository metadata are present
+    def meta_data_present(url, product_dir)
```
jreidinger: maybe call it `meta_data_present?` ?

test/product_location_test.rb
```diff
@@ -0,0 +1,62 @@
...
+
+# loading all repositories and evaluating the product dependencies
+# using the solver takes some time, run it only once and cache
+# the result for all tests
+dir = File.join(__dir__, "data/zypp/test_offline_repo")
+repo_url = "dir://#{URI.escape(dir)}"
+scan_result = Y2Packager::ProductLocation.scan(repo_url)
+
+describe Y2Packager::ProductLocation do
+  describe ".scan" do
```
jreidinger: :+1:  good to have tests for it.

src/lib/y2packager/dialogs/addon_selector.rb
lslezak: Ncurses screenshot added to the PR description. :arrow_up:

src/include/packager/repositories_include.rb
lslezak: Yes, I also noticed that, but that's out of the scope of this PR...

src/lib/y2packager/product_location.rb
lslezak: Well, the problem is that all that data is optional, if the repository does not contain any product then we have no data to use here.

src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -138,7 +138,7 @@ def dialog_content
       end
 
       # select the dependant products for the active selection
```
jreidinger: dependant used also in this comment ( but not in comment below )

src/lib/y2packager/dialogs/addon_selector.rb
lslezak: OK, that will be changed by the Rubocop automatically (or at least detected). So let's fix that later...

src/lib/y2packager/repomd_downloader.rb
lslezak: Renamed, documented. I'd avoid caching as the files might be removed from the outside. If somebody needs caching it must be implemented on the caller side.

src/lib/y2packager/product_location.rb
lslezak: It's not used yet, but later the product sorter already allows `nil`, it's OK.

src/lib/y2packager/product_finder.rb
lslezak: Because we do not exactly know what was the problem. Maybe a missing dependency or conflict or ..??? And it would be strange to ask the user to resolve a conflict *before* adding the repository actually.

src/lib/y2packager/product_finder.rb
```diff
@@ -142,6 +90,88 @@ def display_order(solvable)
       order
     end
 
+    #
+    # Evaluate the products
+    #
+    # @param product_solvable [Solv] the product solvable to create
+    # @param base_products [Array<String>] the found base products
```
jreidinger: params do not match real params

src/lib/y2packager/product_finder.rb
```diff
@@ -142,6 +90,88 @@ def display_order(solvable)
...
+          next if new_solvable == product_solvable
+          new_solvable.lookup_deparray(Solv::SOLVABLE_PROVIDES).each do |dep|
+            next unless dep.str.start_with?("product(")
+            required << new_solvable.repo.name
+          end
+        end
+
+        required.uniq!
+        required.sort!
+      end
```
jreidinger: :+1:  for moving to own method, but what about moving also this computation of requirements to own method that can also return nil. So in the end it can look like `depends_on:      required(product_solvable, selected_base)` and whole first part of method can live there.

src/lib/y2packager/product_location_details.rb
```diff
@@ -0,0 +1,34 @@
...
+    # @return [Boolean] Base product flag (true if this is a base product)
+    attr_reader :base
+    # @return [Array<String>] The product dependencies, includes also the transitive
+    #  (indirect) dependencies
+    attr_reader :depends_on
+
+    # Constructor
+    #
+    def initialize(product: nil, summary: nil, product_package: nil,
+      order: nil, description: nil, base: nil, depends_on: nil)
```
jreidinger: does default value for attributes still makes sense? nil everywhere? e.g. in depends on it makes some sense like if solver failed ( BTW which should be documented ), but in some I think it should not be nil.

src/lib/y2packager/product_location.rb
```diff
@@ -84,29 +72,25 @@ def self.scan(url, base_product = nil)
       finder.products(base_product).map do |p|
         media_name_pair = downloader.product_repos.find { |r| r[1] == p[:prod_dir] }
         media_name = media_name_pair ? media_name_pair.first : p[:prod_dir]
-        new(media_name, p[:prod_dir],
-          product_name: p[:product_name], summary: p[:summary],
-          description: p[:description], base: p[:base], order: p[:order],
-          depends_on: p[:depends_on], product_package: p[:product_package])
+
+        if p[:product_name]
```
jreidinger: :+1: for this new class for details.

src/lib/y2packager/product_finder.rb
```diff
@@ -176,6 +152,40 @@ def select_products(product_solvable, base)
...
+        next if new_solvable == product_solvable
+        new_solvable.lookup_deparray(Solv::SOLVABLE_PROVIDES).each do |dep|
+          next unless dep.str.start_with?("product(")
+          ret << new_solvable.repo.name
+        end
+      end
+
+      ret.uniq!
+      ret.sort
+    end
```
jreidinger: :+1: much better

src/lib/y2packager/product_finder.rb
```diff
@@ -148,10 +124,10 @@ def create_products(product_solvable, found_base_products, selected_base)
           base:         found_base_products.include?(product_name)
         }
 
-        list << ret.merge(product_data)
+        ret << data.merge(product_data)
```
jreidinger: and when I see this data. Why not create that ProductDetails and return just object?

src/lib/y2packager/dialogs/addon_selector.rb
lslezak: I'd keep the current code, this uses one too long line which would need to be split and decreasing the readability...

src/lib/y2packager/product_location_details.rb
lslezak: OK, updated

src/lib/y2packager/product_finder.rb
```diff
@@ -30,17 +31,30 @@ def initialize(pool)
     #
     # @param selected_base [String] The name of the base product used for evaluating the
     #  dependencies.
+    # @param media_names [Array<Array<String,String>>] Product names
     #
     # @return [Array<Hash>] The list of found products
```
jreidinger: I think it is no longer hash ( Which is good, just not documented here ).

PR #460, opened 2019-07-15
Fixed an unit test to work also on non-x86_64 archs

lslezak:
- The testing repository metadata contains only the x86_64 packages.
- Because the solver ignores the incompatible packages the test failed on archs other than x86_64.
- We need to mock the `Solv::Poll.setarch` call to always pass the `x86_64` string to always set the x86_64 architecture during the tests.
- That required a small refactoring in the test and we cannot cache the call (we need to add the mock before running the tests), it's a bit slower now. But not much, on my fast machine it's about from 0.4s to 1.3s, that's acceptable.
- Tested manually with the `osc build openSUSE_Tumbleweed i586` command
- Travis still fails as OBS has not rebuilt the packages yet. :worried:  Works fine locally.

### DIFF
--- a/test/product_location_test.rb
+++ b/test/product_location_test.rb
@@ -10,14 +10,24 @@ def find_product(arr, product)
   arr.find { |p| p.details.product == product }
 end
 
-# loading all repositories and evaluating the product dependencies
-# using the solver takes some time, run it only once and cache
-# the result for all tests
-dir = File.join(__dir__, "data/zypp/test_offline_repo")
-repo_url = "dir://#{URI.escape(dir)}"
-scan_result = Y2Packager::ProductLocation.scan(repo_url)
+# URL of the local testing repository
+REPO_URL = "dir://#{URI.escape(File.join(DATA_PATH, "zypp/test_offline_repo"))}".freeze
 
 describe Y2Packager::ProductLocation do
+  let(:scan_result) { Y2Packager::ProductLocation.scan(REPO_URL) }
+
+  before do
+    # the testing repository only contains the x86_64 packages/products
+    # and the solver ignores the packages for incompatible architectures,
+    # that means the test would fail anywhere except on x86_64.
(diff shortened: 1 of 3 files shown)


test/product_location_test.rb
```diff
@@ -10,14 +10,24 @@ def find_product(arr, product)
...
+
+  before do
+    # the testing repository only contains the x86_64 packages/products
+    # and the solver ignores the packages for incompatible architectures,
+    # that means the test would fail anywhere except on x86_64.
+    #
+    # So we modify the "setarch" call to always pass the "x86_64" parameter.
+    allow_any_instance_of(::Solv::Pool).to receive(:setarch)
+      .and_wrap_original do |method, *_args|
+        method.call("x86_64")
```
jreidinger: another option is to use at https://github.com/yast/yast-packager/blob/8275a2230b9057313f3b62b4465c558b4142fefd/src/lib/y2packager/solvable_pool.rb#L20 Yast::Arch.architecture if it is same as uname -m output see https://github.com/openSUSE/libsolv/blob/master/doc/libsolv-bindings.txt#methods But I worry it won't help with caching. But what can definitivelly helps with caching is use instead of before(:each) to use before(:all)

so it can looks like
```
before(:all) do
  allow_any_instance_of(::Solv::Pool).to receive(:setarch)
      .and_wrap_original do |method, *_args|
        method.call("x86_64")
end

def scan_result
  @scan_result ||= Y2Packager::ProductLocation.scan(REPO_URL)
end
```

this way it will be cached.

lslezak: Actually it does not work, the `before(:all)` causes error "The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported."

And for some reason the `@scan_result` is nil for each, which means the cache does not work.

So I'd leave it as it is now.

PR #461, opened 2019-07-23
Add #summary method

joseivanlopez:
Method `ProductLocation#summary` was missing.

https://bugzilla.suse.com/show_bug.cgi?id=1142414

Related to: https://github.com/yast/yast-packager/pull/459

### DIFF
--- a/src/lib/y2packager/product_location.rb
+++ b/src/lib/y2packager/product_location.rb
@@ -83,7 +83,7 @@ module Y2Packager
     #
     # @return [String, nil] nil if there is no details or details does not contain a summary
     def summary
-      return nil if details.nil? || details.summary.empty?
+      return nil if !details&.summary || details.summary.empty?
 
       details.summary
     end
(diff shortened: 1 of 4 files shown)


src/lib/y2packager/product_location.rb
```diff
@@ -78,5 +78,14 @@ def initialize(name, dir, product: nil)
       @dir = dir
       @details = product
     end
+
+    # Product summary taken from its details
+    #
+    # @return [String, nil] nil if there is no details or details does not contain a summary
+    def summary
+      return nil if details.nil? || details.summary.empty?
```
lslezak: This can be simplified to
```suggestion
      return nil if details&.summary&.empty?
```

but I'm not sure if it is better readable. What do you think?

joseivanlopez: That check is not exactly the same, because it does not return when either `details` or `summary` is `nil`.

Umm, the correct replacement would be:

```
return nil if !details&.summary || details.summary.empty?
```

I will use this, because current version does not take into account when `summary` is nil (although it should not be nil).

PR #462, opened 2019-07-24
Add missing #depends_on method to ProductLocation

dgdavid:
## Problem

The `ProductLocation` does not have the `#depends_on` method that is trying to be used in the [`addon_selector` dialog](https://github.com/yast/yast-packager/blob/79d45133b84d6afa0de9afe95e816dfd66d85731/src/lib/y2packager/dialogs/addon_selector.rb#L222). Actually, it is a method of its details (`ProductLocationDetails`).

* https://bugzilla.suse.com/show_bug.cgi?id=1142535
* https://trello.com/c/ItpNy9io/1156-sles15-sp2-p5-1142535-installataion-fails-in-addonproductsle-tests-for-sle-15-sp2-addonselectorrb222

## Solution 

Add the missing method to `ProductLocation`.

**Quite similar** to what was done in #461 and also related with changes introduced in ~~#84~~ #459

### DIFF
--- a/src/lib/y2packager/product_location.rb
+++ b/src/lib/y2packager/product_location.rb
@@ -87,12 +87,5 @@ module Y2Packager
 
       details.summary
     end
-
-    # Product dependencies taken from its details
-    #
-    # @return [Array] empty list if there is no dependencies or the dependencies list
-    def depends_on
-      details&.depends_on || []
-    end
   end
 end
--- a/src/lib/y2packager/product_location.rb
+++ b/src/lib/y2packager/product_location.rb
@@ -87,5 +87,12 @@ module Y2Packager
 
       details.summary
     end
+
+    # Product dependencies taken from its details
+    #
+    # @return [Array] empty list if there is no dependencies or the dependencies list
+    def depends_on
+      details&.depends_on || []
+    end
   end
 end
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Jul 24 14:21:07 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Add method to get the dependencies of a product (bsc#1142535).
+- 4.2.20
+
+-------------------------------------------------------------------
 Tue Jul 23 10:25:51 UTC 2019 - José Iván López González <jlopez@suse.com>
 
 - Add method to get the summary of a product (bsc#1142414).
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.2.19
+Version:        4.2.20
 Release:        0
 Summary:        YaST2 - Package Library
 License:        GPL-2.0-or-later
--- a/src/lib/y2packager/dialogs/addon_selector.rb
+++ b/src/lib/y2packager/dialogs/addon_selector.rb
@@ -149,7 +149,8 @@ module Y2Packager
         new_items.each do |p|
           # the dependencies contain also the transitive (indirect) dependencies,
           # we do not need to recursively evaluate the list
-          selected_items.concat(p.depends_on)
+          dependencies = p&.details&.depends_on
+          selected_items.concat(dependencies) if dependencies
         end
 
         selected_items.uniq!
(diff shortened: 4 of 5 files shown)


src/lib/y2packager/product_location.rb
```diff
@@ -87,5 +87,12 @@ def summary
 
       details.summary
     end
+
+    # Product dependencies taken from its details
+    #
+    # @return [Array] empty list if there is no dependencies or the dependencies list
+    def depends_on
+      details&.depends_on || []
```
lslezak: Hm, I'd rather fix this line https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/dialogs/addon_selector.rb#L222 instead adding a wrapper here. In the end we would add wrappers for all methods and that would defeat the purpose of having a separate `ProductLocationDetails` class...

dgdavid: That's was a doubt that prompt to my mind while I was solving the issue. And also I think that the `ProductLocationDetails#initialize` should ensure to _sanitize_ the `depends_on` param, because is expected that it will be always an array, isn't it? (see https://github.com/yast/yast-packager/blob/79d45133b84d6afa0de9afe95e816dfd66d85731/src/lib/y2packager/dialogs/addon_selector.rb#L152-L156)

lslezak: The `nil` value for the `depends_on` is used to indicate that the were dependency issues and the dependencies could not have been computed.

Currently that's is not used in the UI, but in theory we could display some error/warning message instead of an empty list.

joseivanlopez: The problem I see with the `ProductLocation` and `ProductLocationDetails` approach is that the details attribute could be nil. This happens because `:product` is an optional parameter in `ProductLocaltion#initialize`.

So, checking the existence of details is always mandatory: `product.details&.attribute`. And this has an implicit drawback (IMHO). The caller is the responsible to define the default value when details or the attribute are nil:

```
product.details&.attribute || []
```

For me, a better solution would be to force `ProductLocation` to always have a details object. And then I would define proper default values for each detail attribute:

```
class ProductLocation
  def initialize(name, dir, product: nil)
    @details = product || ProductLocationDetails.new
  end
end

class ProductLocaltionDetails
  def initialize(product: nil, depends_on: [], ...)
  end
end
```

dgdavid: >  And then I would define proper default values for each detail attribute

But in the previous comment, @lslezak said

>  The nil value for the depends_on is used to indicate that the were dependency issues and the dependencies could not have been computed.And then I would define proper default values 

But I still thinking that this will be problematic, since

```irb
irb(main):001:0> [].concat(nil)
        5: from /usr/bin/irb:23:in `<main>'
        4: from /usr/bin/irb:23:in `load'
        3: from /usr/lib64/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):1
        1: from (irb):1:in `concat'
TypeError (no implicit conversion of nil into Array)
irb(main):002:0> 
```

joseivanlopez: Yes, I don't know all the design decisions behind this API. But I would use another mechanism to indicate dependency issues. I would keep `#depends_on` as an empty list even when there are issues with dependencies (this avoid to check for nil everywhere), and also I would extend the API with an specific method to ask for possible issues: `#dependency_errors?`.  But, again, this is only my personal feeling, Ladislav knows the details much better.

lslezak: > For me, a better solution would be to force `ProductLocation` to always have a details object.

In some cases the repository might not contain a product (e.g. a custom or a 3rd party repository like Packman). In that case we do not have any details, we can only read the data from the `/media.1/products` file which only contains directory name and add-on name.

In that case it would make sense to have `nil` details

dgdavid: @lslezak, @joseivanlopez, 

All things considered, I'll fix the issue in https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/dialogs/addon_selector.rb#L222, although I have a similar feeling than Iván.

lslezak: OK, I'd prefer a quick/simple fix to avoid crashes, we can discuss and improve it later...

PR #463, opened 2019-07-25
Set the proper release version always

dgdavid:
## Problem

> #TODO: explain the problem

* https://bugzilla.suse.com/show_bug.cgi?id=1132748
## Solution

Set the `RELEASEVER_ENV` environment variable **always**, not only during the upgrade.

## See also

[This comment](https://github.com/yast/yast-packager/pull/383#pullrequestreview-170590620) in #383,  where this _feature_ was introduced only for the `update` mode.

### DIFF
--- a/src/lib/y2packager/clients/inst_productsources.rb
+++ b/src/lib/y2packager/clients/inst_productsources.rb
@@ -140,9 +140,9 @@ module Yast
         return :auto
       end
 
-      # set release version for update as for newly added repositories
+      # Set the proper release version for newly added repositories
       # we want to use new product and not old
-      ENV[RELEASEVER_ENV] = Product.version if Mode.update
+      ENV[RELEASEVER_ENV] = Product.version
 
       # (Applicable only in inst-sys)
       @preselect_recommended = true
--- a/test/lib/clients/inst_productsources_test.rb
+++ b/test/lib/clients/inst_productsources_test.rb
@@ -17,6 +17,14 @@ describe Yast::InstProductsourcesClient do
       allow(Yast::Product).to receive(:version).and_return(product_version)
     end
 
+    around do |test|
+      previous_env = ENV.to_hash
+
+      test.run
+
+      ENV.replace(previous_env)
+    end
+
     it "returns :auto if AddOnProduct is set to skip" do
       allow(Yast::AddOnProduct).to receive(:skip_add_ons).and_return(true)
 
--- a/test/lib/clients/inst_productsources_test.rb
+++ b/test/lib/clients/inst_productsources_test.rb
@@ -50,14 +50,10 @@ describe Yast::InstProductsourcesClient do
       client.main
     end
 
-    it "sets the RELEASEVER_ENV to Product.version" do
-      stub_const("#{described_class}::RELEASEVER_ENV", "ZYPP_RELEASEVER_ENV")
-
-      expect(ENV[described_class::RELEASEVER_ENV]).to be_nil
+    it "sets the ZYPP_REPO_RELEASEVER environment variable to Product.version" do
+      expect(ENV).to receive(:[]=).with("ZYPP_REPO_RELEASEVER", "15.1")
 
       client.main
-
-      expect(ENV[described_class::RELEASEVER_ENV]).to eq(product_version)
     end
   end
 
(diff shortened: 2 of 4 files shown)


src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -140,9 +140,9 @@ def main
         return :auto
       end
 
-      # set release version for update as for newly added repositories
+      # Set the proper release version for newly added repositories
       # we want to use new product and not old
```
dgdavid: Maybe this comment could be updated/improved, but right now I have no inspiration :) Do you have any idea? Thanks!

test/lib/clients/inst_productsources_test.rb
```diff
@@ -46,6 +49,16 @@
...
     end
+
+    it "sets the RELEASEVER_ENV to Product.version" do
+      stub_const("#{described_class}::RELEASEVER_ENV", "ZYPP_RELEASEVER_ENV")
+
+      expect(ENV[described_class::RELEASEVER_ENV]).to be_nil
+
+      client.main
+
+      expect(ENV[described_class::RELEASEVER_ENV]).to eq(product_version)
```
lslezak: This test actually really modifies the environment... :worried: 

What about stubbing the change like this:

```ruby
expect(ENV).to receive(:[]=).with(...)
```

The modified the ENV could leak to another tests and cause random failures...

test/lib/clients/inst_productsources_test.rb
```diff
@@ -46,6 +49,16 @@
 
       client.main
     end
+
+    it "sets the RELEASEVER_ENV to Product.version" do
+      stub_const("#{described_class}::RELEASEVER_ENV", "ZYPP_RELEASEVER_ENV")
```
jreidinger: is this needed? why stubbing it?

dgdavid: Hmmm, probably not. Re-reading my own test now, I realize that the important part is that the code set the "ZYPP_RELEASEVER_ENV" no matter how we called/stored it internally :)

Thanks!

test/lib/clients/inst_productsources_test.rb
dgdavid: Intersting, because it was my first idea... but I though in that way I could not test the key, only the value. But I did not test it :sweat_smile: 

Now, I can see that I can do it, simply given two params to `.with` :sweat_smile: :sweat_smile: 

Thanks!

test/lib/clients/inst_productsources_test.rb
```diff
@@ -46,6 +49,12 @@
 
       client.main
     end
+
+    it "sets the ZYPP_REPO_RELEASEVER environment variable to Product.version" do
+      expect(ENV).to receive(:[]=).with("ZYPP_REPO_RELEASEVER", "15.1")
```
jreidinger: well, issue is that also other tests modify that env.
so I suggest to add something like
```
around do |test|
      old_env = ENV.to_hash
      test.call
      ENV.replace(old_env)
end
```

I use e.g. also in cheetah https://github.com/openSUSE/cheetah/blob/master/lib/cheetah.rb#L416

test/lib/clients/inst_productsources_test.rb
```diff
@@ -7,11 +7,22 @@
...
+
   describe "#main" do
     before do
       allow(Yast::Sequencer).to receive(:Run)
       allow(Yast::Wizard).to receive(:OpenDialog)
       allow(Yast::Wizard).to receive(:CloseDialog)
+      allow(Yast::Product).to receive(:version).and_return(product_version)
+    end
+
+    around do |test|
```
lslezak: Do we still need this "around" block when we mock `ENV[]=` ?

jreidinger: yes, as other tests run also main, which sets ENV

test/lib/clients/inst_productsources_test.rb
jreidinger: BTW you no longer need to mock this and can simply check ENV as it was before.

test/lib/clients/inst_productsources_test.rb
dgdavid: As we discussed in IRC, we will keep the `around` hook because 

```
<jreidinger> [...] all other examples also run client.main
<jreidinger> dgdavid: so it modifies ENV of whole test and can have consequences
```

PR #466, opened 2019-08-02
Fix the forced base product selection (2/3)

dgdavid:
## Problem

Recently, the `select_product` option was added to the control file to allow **forcing** the base product selection when having multiple products in a single repository (see https://github.com/yast/yast-installation/pull/790).

But seems that the feature was not fully implemented and the installer only avoids displaying the product selector, but not marking the forced product as selected, raising an internal error when trying to access to `#installation_package`.

* https://bugzilla.suse.com/show_bug.cgi?id=1143943
* https://trello.com/c/YuiJfYBt/1201-ostumbleweed-p1-1143943-build-20190801-undefined-method-installationpackage

## Solution

To select the **foced** product properly.

:warning: **Requires `yast2 4.2.17` - https://github.com/yast/yast-yast2/pull/954**

## Additional notes

See https://github.com/yast/yast-yast2/pull/954 for more information and screenshots.

### DIFF
--- a/src/lib/y2packager/clients/inst_repositories_initialization.rb
+++ b/src/lib/y2packager/clients/inst_repositories_initialization.rb
@@ -88,7 +88,14 @@ module Y2Packager
       # @see https://github.com/yast/yast-packager/blob/7e1a0bbb90823b03c15d92f408036a560dca8aa3/src/modules/Packages.rb#L1876
       # @see https://github.com/yast/yast-packager/blob/fbc396df910e297915f9f785fc460e72e30d1948/src/modules/Packages.rb#L1905
       def adjust_base_product_selection
-        if products.size == 1
+        if forced_base_product
+          log.info("control.xml wants to force the #{forced_base_product.name} product")
+
+          forced_base_product.select
+          discarded_products = products.reject { |p| p == forced_base_product }
+
+          log.info("Ignoring the other products: #{discarded_products.map(&:name).join(", ")}")
+        elsif products.size == 1
           products.first.select
         else
           products.each(&:restore)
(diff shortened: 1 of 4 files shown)


src/lib/y2packager/clients/inst_repositories_initialization.rb
```diff
@@ -88,7 +88,14 @@ def init_installation_repositories
...
       # @see https://github.com/yast/yast-packager/blob/fbc396df910e297915f9f785fc460e72e30d1948/src/modules/Packages.rb#L1905
       def adjust_base_product_selection
-        if products.size == 1
+        if forced_base_product
+          log.info("control.xml wants to force the #{forced_base_product.name} product")
+
+          forced_base_product.select
+          discarded_products = products.reject { |p| p == forced_base_product }
+
+          log.info("Ignoring the other products: #{discarded_products.map(&:name).join(", ")}")
```
lslezak: For logged messages simple `inspect` is better IMHO as it better represents the internal value.

```suggestion
          log.info("Ignoring the other products: #{discarded_products.map(&:name).inspect}")
```

src/lib/y2packager/clients/inst_repositories_initialization.rb
```diff
@@ -101,6 +108,13 @@ def adjust_base_product_selection
       def products
         @products ||= Y2Packager::Product.available_base_products
       end
+
+      # Return the forced base product (if any)
+      #
+      # @return [Y2Product, nil]
+      def forced_base_product
+        @forced_base_product ||= Y2Packager::Product.forced_base_product
```
lslezak: The forced base product is already cached here
https://github.com/yast/yast-yast2/pull/954/files#diff-f69b205c192a907ae0fa6ae4feebaab1R125

Do we need to cache it again here? In the code above it would be enough to just use  
`forced_base_product = Y2Packager::Product.forced_base_product`

dgdavid: You're right, I overlooked it :)

Thanks!

PR #467, opened 2019-08-05
Pattern deselection (SP5)

lslezak:
## Problem

- After deselecting the X11 and the GNOME patterns (in this order!) YaST selects the X11 pattern back when confirming the changes and going back to the installation proposal.
- The YaST code already does not select again the patterns which have been unselected byt the user. However, the problem is that the solver looses the information that the X11 pattern was unselected by the user, it becomes unselected by the solver (because of the GNOME -> X11 dependency).

pattern_deselection

pattern_deselection_broken

## Fix

- Because the package manager is reset at some point we have to call the pattern selection again. (My first attempt was to select the patterns only once and dot select them again. But that did not work well, in the end no pattern was selected.)
- To avoid selecting the unselected patterns we need to remember the unselected patterns when leaving the package manager UI. Later we do not select them again.
- Tested manually, after unselecting both patterns the X11 pattern is not selected again and the other default patterns are correctly selected.

pattern_deselection_fixed

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Aug  5 18:36:42 UTC 2019 - Ladislav Slezák <lslezak@suse.cz>
+
+- Better handle unselecting the default pre-selected patterns
+  by the user (bsc#1140735)
+- 3.3.2
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


lslezak:
## Note

We do not need to fix it in the `master` branch, the `Pkg.PkgApplReset` call is not used there, the patterns are not selected again. The problem does not happen there.

PR #468, opened 2019-08-28
Fix message error

joseivanlopez:
### Problem

An exception is raised when there are AutoYaST errors. The problem is that frozen strings from translated error messages are modified.

* https://bugzilla.suse.com/show_bug.cgi?id=1148536

### Solution

Avoid modifications of frozen strings.

### Testing

* Added unit test.
* Manually tested


joseivanlopez:
I think this bug is there since SLE-15-GA: https://github.com/yast/yast-packager/pull/301. Maybe we should backport it, but the bug has not been reported yet (only detected by me in SLE-15-SP2).

PR #472, opened 2019-09-04
Detect Installation Medium Type

lslezak:
- This is a helper which will be used the code detect the online/offline/standard medium and handle the different media specifically.
- Related to https://jira.suse.com/browse/SLE-7214

### DIFF
--- a/src/lib/y2packager/medium_type.rb
+++ b/src/lib/y2packager/medium_type.rb
@@ -21,87 +21,87 @@ module Y2Packager
   # This class scans the installation medium type and detects which installation
   # medium type it is (online/offline/standard).
   class MediumType
-    extend Yast::Logger
+    class << self
+      include Yast::Logger
 
-    # Type of the installation medium, raises an exception if the installation
-    # URL is not set (nil) or is empty.
-    #
-    # @return [Symbol] Symbol describing the medium, one of `:offline`,
-    # `:online` or `:standard`
-    def self.type
-      @type ||= detect_medium_type
-    end
-
-    # Is the medium an online installation medium? (SLE Online)
-    # Raises an exception if the installation URL is not set (nil) or is empty.
-    # The online installation medium contains no repository
-    # or a repository without any base product.
-    def self.online?
-      type == :online
-    end
-
-    # Is the medium an offline installation medium?  (SLE Offline)
-    # Raises an exception if the installation URL is not set (nil) or is empty.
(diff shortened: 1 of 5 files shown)


src/lib/y2packager/medium_type.rb
```diff
@@ -0,0 +1,107 @@
...
+require "y2packager/product_location"
+require "y2packager/repomd_downloader"
+
+Yast.import "InstURL"
+
+module Y2Packager
+  # This class scans the installation medium type and detects which installation
+  # medium type it is (online/offline/standard).
+  class MediumType
+    extend Yast::Logger
```
jreidinger: hmm, whole class is class based, so what about simple `class << self` and then common include and also private?

src/lib/y2packager/medium_type.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    # Type of the installation medium, raises an exception if the installation
+    # URL is not set (nil) or is empty.
+    #
+    # @return [Symbol] Symbol describing the medium, one of `:offline`,
+    # `:online` or `:standard`
+    def self.type
+      @type ||= detect_medium_type
```
jreidinger: hmm, this is quite confusing. What about returning instance of medium type, which has methods like `online?` and also name of this method I expect something like `detect` or `read` so it detect type.

lslezak: I'm not sure. I do not like the `detect` name because the detection is done only once and then the result is cached, this might be confusing. And I also want to avoid a long method chain like `Y2Packager::MediumType.detect.online?`, that's why there is the `Y2Packager::MediumType.online?` shortcut.

I introduced the `type` method just to make it easy to use it in `switch` although I expect that the most of the code will use just the `online?`/`offline?` methods.

jreidinger: well, it can be better in situation when you want to store medium type like
```
medium_type = Y2Packager::MediumType.detect
if medium_type.online?
```

it also allows introspect and allows in future smarter methods in medium type, like "slow?` or `need_additional_ram?`

but it is up to you.

lslezak: OK, I see your point now. But it's more about the medium layout than about the physical properties of the medium. In this case we need to know if the medium provides packages for full installation (or at least for some minimal system) or it's just a boot medium and the packages are downloaded from SCC.

I'd keep it simple.

Thanks!

PR #475, opened 2019-09-27
Support for the offline installation medium (jsc#SLE-7101)

lslezak:
- https://jira.suse.com/browse/SLE-7101
- Small adaptations for the offline medium installer
- `!ENV[] == "1"` actually means `(!ENV[]) == "1"` which always evaluates to `false` :open_mouth:
- 4.2.27

### DIFF
--- a/src/lib/y2packager/dialogs/addon_selector.rb
+++ b/src/lib/y2packager/dialogs/addon_selector.rb
@@ -45,7 +45,7 @@ module Y2Packager
         @products = products
         # do not offer base products, they would conflict with the already selected base product,
         # allow a hidden way to force displaying them in some special cases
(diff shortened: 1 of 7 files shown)


src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -45,7 +45,7 @@ def initialize(products)
         @products = products
         # do not offer base products, they would conflict with the already selected base product,
         # allow a hidden way to force displaying them in some special cases
-        @products.reject! { |p| p.details&.base } if !ENV["Y2_DISPLAY_BASE_PRODUCTS"] == "1"
+        @products.reject! { |p| p.details&.base } unless ENV["Y2_DISPLAY_BASE_PRODUCTS"] == "1"
```
jreidinger: or maybe even easier to `if ENV["Y2_DISPLAY_BASE_PRODUCTS"] != "1"`

PR #477, opened 2019-10-01
Do not add an empty repository from the offline medium root  (jsc#SLE-7101)

lslezak:
- Workaround to avoid adding an empty repository from the root of the offline medium
- https://jira.suse.com/browse/SLE-7101
- 4.2.30

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -7,6 +7,7 @@ require "uri"
 require "cgi"
 require "shellwords"
 
+require "y2packager/medium_type"
 require "y2packager/product_upgrade"
 require "y2packager/resolvable"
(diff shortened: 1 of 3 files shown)


src/modules/Packages.rb
```diff
@@ -1725,6 +1726,11 @@ def Initialize_StageInitial(show_popup, base_url, log_url, product_dir = "/")
       if !meta_data_present?(base_url, product_dir)
         log.info "Metadata not found at #{log_url}, postponing the repository initialization"
         return
+      # TODO: the offline medium contains and empty repository in the root,
```
jreidinger: typo "and" instead of "an"

PR #482, opened 2019-10-22
Include some hints in the "Priority" label (bsc#1154693)

lslezak:
## The Problem

- The meaning of the "Priority" field in the repository manager is a bit misleading
- The higher number the lower priority - that's quite unexpected and confusing (that was inherited from the yum package manager)
- https://bugzilla.suse.com/show_bug.cgi?id=1154693

## Solution

- Add a hint to the widget label if the terminal is wide enough (so there is space for it)
- I was thinking what to use there, the longer version is used only on larger terminals, but a too long label still might not fit.
- One option was to use something like "the higher number the lower priority" but then I was afraid that some translations would be too long. So I decided to use "higher number = lower priority" which is grammatically not that nice but indicates to the translators that they should use some shorter version.

## Screenshots

*(I deliberately installed the French translations as they are usually longer.)*

Small window (80x25), the short label is used (cough, cough, it already fits so so...):
repo_priority_short

Larger window, the new longer label is used (obviously not translated yet):
repo_priority_long

In GUI it always uses the longer version, there should be almost always enough space (this is in  800x600):
repo_priority_qt

- 4.2.31


lslezak:
@mvidner What do you think about this label? Is it OK? Or any better suggestion?

mvidner:
It's hard, I don't see a way to make it obvious and unmistakable. Let's see how the world sees "priority"
- https://ux.stackexchange.com/questions/73445/is-a-higher-priority-a-smaller-number
- https://en.wikipedia.org/wiki/Priority

mvidner:
You have missed an important aspect of the bug report: the priority changing widget is an IntField which has **arrows**: an up arrow raises the number, which lowers the priority. That contributes to the confusion.
IMHO we should replace the widget with an editable combo box, with values
- "1 (high)"
- "50 (medium)"
- "99 (low, default)"

lslezak:
An editable combo box looks like an interesting alternative... I'll check that later.

rindeal:
How about renaming "Priority" field to "Weight"?

mvidner:
BTW on Leap 15.2, "zypper addrepo ..." tells me

> Repository priorities are without effect. All enabled repositories share the same priority.

mvidner:
Ah, now I understand!

I have interpreted "Repository priorities are without effect." as "the priority feature is gone, zypp ignores whatever priority you specify", but the [source code](https://github.com/openSUSE/zypper/blob/0c0003fb6685c5216f17375fcc0948ef01da0048/src/repos.cc#L179) and the following example show that it in fact means "the currently enabled repositories happen to share the same priority, so I am confusing you with superfluous information. If you add a repo with a different priority I will remind you too which will be much more useful."

```console
# LANG=C z ar -p 50 /tmp tmp50
Adding repository 'tmp50' ...............................................[done]
Repository 'tmp50' successfully added

URI         : dir:/tmp
Enabled     : Yes
GPG Check   : Yes
Autorefresh : No
Priority    : 50 (raised priority)

Repository priorities in effect:                (See 'zypper lr -P' for details)
      50 (raised priority)  :  1 repository
      99 (default priority) : 13 repositories
```

dgdavid:
@lslezak, @mvidner,

What is the status of this PR? Despite the conflicting files, it looks almost finished, am I wrong?

lslezak:
Um, it's basically just waiting for the @mvidner 's approval.

mvidner (approved):
Thanks, this is an easy improvement

PR #484, opened 2019-10-30
Aligned lang of license text with the selected language (bsc#1152482)

cwh42:
Bug has been introduced because the license did not respect the LANG env variable. Unfortunately after that it *only* followed LANG.
That was introduced 10 years ago and no one noticed.

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -863,7 +863,7 @@ module Yast
       # by now (see bnc#504803, c#28). Language::Language does only
       # sysconfig reading, which is not too useful in cases like
       # 'LANG=foo_BAR yast repositories'
-      language = EnvLangToLangCode(Builtins.getenv("LANG"))
+      language = Language.language || EnvLangToLangCode(ENV["LANG"])
 
       # Preferencies how the client selects from available languages
       langs = [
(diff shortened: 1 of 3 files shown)


lslezak:
Travis failed because of an outdated version of pkg-bindings. This was caused by renaming the base images so all builds since that failed. Fixed in https://github.com/yast/docker-yast-ruby/pull/36, the future builds should be OK.

src/modules/ProductLicense.rb
```diff
@@ -863,7 +863,7 @@ def InitLicenseData(src_id, dir, licenses, available_langs, require_agreement, l
       # by now (see bnc#504803, c#28). Language::Language does only
       # sysconfig reading, which is not too useful in cases like
       # 'LANG=foo_BAR yast repositories'
-      language = EnvLangToLangCode(Builtins.getenv("LANG"))
+      language = Language.language || EnvLangToLangCode(ENV["LANG"])
```
mvidner: @cwh42 Um, this review is late, but I think this code would only work if `Language.language` were `nil`. But it's initialized to "en_US" in the constructor: https://github.com/yast/yast-country/blob/SLE-12-SP5/language/src/modules/Language.rb#L65

cwh42: According to the comment right above that line of code how it was before my change has been introduced to fix that language in some cases was `nil`. The result was, that `language` was **only** set to whatever was defined in LANG while it was impossible to change it by changing `Language.language`. I tested that it works now and also works when LANG is used to set the language - which is the case when language is selected in the bootloader.

mvidner: But that's just theorizing. Your tests show it works.

PR #485, opened 2019-10-31
 Decreasing required memory

schubi2:
Replacing old Pkg.ResolvableProperties()and Pkg.ResolvabeDependencies() calls by Y2Packager::Resolvable.any? and Y2Packager::Resolvable.find in order to save memory.

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -24,9 +24,6 @@ module Yast
     # All known types of resolvables
     RESOLVABLE_TYPES = [:product, :patch, :package, :pattern, :language].freeze
 
-    # Key to sort by resolvable selection
-    RESOLVABLE_SORT_ORDER = { product: "source", pattern: "order" }.freeze
-
     # Minimum set of packages tags required to enable VNC server
     VNC_BASE_TAGS = ["xorg-x11", "xorg-x11-Xvnc", "xorg-x11-fonts"].freeze
     # Additional packages tags needed to run second stage in graphical mode
@@ -170,11 +167,11 @@ module Yast
     # @param [String] format string format string to print summaries in
     # @return a list of selected resolvables
     def ListSelected(what, format)
-      selected = Pkg.ResolvableProperties("", what, "")
+      selected = Y2Packager::Resolvable.find(kind: what)
 
-      selected.select! { |r| r["status"] == :selected }
+      selected.select! { |r| r.status == :selected }
 
-      selected.select! { |r| r["user_visible"] } if what == :pattern
+      selected.select! { |r| r.user_visible } if what == :pattern
 
       sort_resolvable!(selected, what)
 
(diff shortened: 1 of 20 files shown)


src/modules/Packages.rb
```diff
@@ -170,11 +167,11 @@ def ResetProposalCache
     # @param [String] format string format string to print summaries in
     # @return a list of selected resolvables
     def ListSelected(what, format)
-      selected = Pkg.ResolvableProperties("", what, "")
+      selected = Y2Packager::Resolvable.find(kind: what)
 
-      selected.select! { |r| r["status"] == :selected }
+      selected.select! { |r| r.status == :selected }
```
lslezak: Move it to the Resolvable.find call.

src/modules/Packages.rb
```diff
@@ -429,86 +426,9 @@ def CheckDiskSize(init)
     # Checks which products have been selected for removal and modifies
     # the warning messages accordingly.
     #
-    # @param [Yast::ArgRef] ret reference to map MakeProposal->Summary
-    def CheckOldAddOns(ret)
-      products = Pkg.ResolvableProperties("", :product, "")
```
lslezak: Just a question: why we are dropping this?

src/modules/Packages.rb
```diff
@@ -1612,12 +1532,12 @@ def AdjustSourcePropertiesAccordingToProduct(src_id)
...
 
       Builtins.y2milestone("Trying to get repository name from products")
-      all_products = Pkg.ResolvableProperties("", :product, "")
+      all_products = Y2Packager::Resolvable.find(kind: :product)
       new_name = nil
       Builtins.foreach(all_products) do |one_product|
         # source ID matches
-        if Ops.get_integer(one_product, "source", -1) == src_id
-          name = one_product["name"] || ""
+        if (one_product.source || -1) == src_id
```
lslezak: You can remove these defaults.

src/modules/ProductLicense.rb
```diff
@@ -1702,9 +1704,9 @@ def product_license(id, tmpdir)
       Pkg.SourceLoad
       ::FileUtils.mkdir_p(tmpdir)
 
-      products = Pkg.ResolvableProperties("", :product, "")
-      products.select! { |p| p["source"] == id }
-      product_names = products.map { |p| p["name"] }.uniq
+      products = Y2Packager::Resolvable.find(kind: :product)
+      products.select! { |p| p.source == id }
```
lslezak: Move it to the Resolvable.find call.

src/modules/SourceManager.rb
```diff
@@ -580,7 +581,7 @@ def AddUpdateSources(urls)
           ret = Builtins.add(ret, url) # #180820#c26
 
           # is there any patch available?
-          patches = Pkg.ResolvableProperties("", :patch, "")
+          patches = Y2Packager::Resolvable.find(kind: :patch)
```
lslezak: Just use `Resolvable.any?` here, we just detect if there is any patch available at all.

src/modules/Packages.rb
schubi2: We have talked together about it some weeks ago :-)
Currently we do not have any workflow for that.

src/modules/SourceManager.rb
schubi2: But we are also logging how many patches are available.

PR #486, opened 2019-11-01
Do not crash when the product licenses cannot be read (bsc#1155454)

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1155454
- This is kind of unusual bug, the tester used the http://download.suse.de/full/full-sle15-sp2-x86_64/ repository which contains all packages including the SLE and openSUSE products. Some product is broken there as libzypp reports this error:  
```
2019-10-29 11:47:43 <2> install(3486) [zypp] Product.cc(referencePackage):112
    (785)product:--0.noarch(full-sle15-sp2-x86_64): no reference package found:
    product() == -0
```
- Tested manually, now with the fix it does not crash
- 4.2.31

### DIFF
--- a/src/modules/ProductLicense.rb
+++ b/src/modules/ProductLicense.rb
@@ -1711,6 +1711,7 @@ module Yast
       product_names.each do |product_name|
         locales = Pkg.PrdLicenseLocales(product_name)
         log.info("License locales for product #{product_name.inspect}: #{locales.inspect}")
+        next unless locales
 
         locales.each do |locale|
           license = Pkg.PrdGetLicenseToConfirm(product_name, locale)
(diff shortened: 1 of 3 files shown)


src/modules/ProductLicense.rb
```diff
@@ -1711,6 +1711,7 @@ def product_license(id, tmpdir)
       product_names.each do |product_name|
         locales = Pkg.PrdLicenseLocales(product_name)
         log.info("License locales for product #{product_name.inspect}: #{locales.inspect}")
+        next unless locales
```
jreidinger: another option could be simple `locales ||= []` which will behave same.

But in either case please document why this check is needed and ideally also with bug report reference.

PR #487, opened 2019-11-08
Preselect the installed add-ons (jsc#SLE-7101)

lslezak:
- ...when upgrading from the Full installation medium
- Related to https://jira.suse.com/browse/SLE-7101
- Handles also the product renames
- Tested manually
- 4.2.32

## Screenshot

When upgrading from the SP1 the already installed modules are automatically preselected:

preselected_addons

### DIFF
--- a/src/lib/y2packager/dialogs/addon_selector.rb
+++ b/src/lib/y2packager/dialogs/addon_selector.rb
@@ -240,29 +240,10 @@ module Y2Packager
       def preselected_products
         return [] unless Yast::Mode.update
 
-        installed_addons = Y2Packager::Resolvable.find(
-          kind: :product, status: :installed, type: "add-on"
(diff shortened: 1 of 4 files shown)


src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -229,6 +233,38 @@ def product_description(product)
         # render the ERB template in the context of this object
         erb.result(binding)
       end
+
+      # return a list of the preselected products
+      # during upgrade we want to preselect the installed products
+      # @return [Array<Y2Packager::ProductLocation>] the products
+      def preselected_products
```
jreidinger: code looks fine, but can you please find a better location for this piece of code? As I would like to use it also in autoupgrade to select addons to upgrade, so I prefer to call it from some non-dialog location.

PR #488, opened 2019-11-26
Show beta warning only once

imobachgs:
## The problem

According to [bsc#1156629](https://bugzilla.suse.com/show_bug.cgi?id=1156629), the installer shows a beta warning for each beta module. Ideally, it should display the (same) beta warning only once. Of course, if two beta warnings are different, then it should show both of them.

Trello: https://trello.com/c/d8jtS3dA/
## The solution

This PR extends the `InstShowInfo` module to keep track of which information has already been displayed based on the file content (nor on the path). It is something similar to what YaST [already does](https://github.com/yast/yast-yast2/blob/22d33428525ecffbcab4c1cc9a911dce43f07704/library/packages/src/lib/y2packager/license.rb#L124) with licenses.

## Where is InstShowInfo.show_info_txt used?

```txt
registration/src/lib/registration/ui/addon_eula_dialog.rb
144:        Yast::InstShowInfo.show_info_txt(info_file) if File.exist?(info_file)

packager/src/modules/ProductLicense.rb
241:      InstShowInfo.show_info_txt(@beta_file) if !@beta_file.nil?
370:        InstShowInfo.show_info_txt(@beta_file) if !@beta_file.nil?
807:        InstShowInfo.show_info_txt(@beta_file)

installation/src/lib/installation/clients/inst_complex_welcome.rb
67:        InstShowInfo.show_info_txt(BETA_FILE)
```

### DIFF
--- a/src/modules/InstShowInfo.rb
+++ b/src/modules/InstShowInfo.rb
@@ -22,7 +22,7 @@ module Yast
     def show_info_txt(file_path)
       info_file = Y2Packager::InfoFile.read(file_path)
       if info_file.nil?
-        Builtins.y2milestone("No %1", info_file)
+        Builtins.y2milestone("No %1", file_path)
         return
       end
 
--- a/src/modules/InstShowInfo.rb
+++ b/src/modules/InstShowInfo.rb
@@ -1,4 +1,5 @@
 require "yast"
+require "y2packager/info_file"
 
 # Yast namespace
 module Yast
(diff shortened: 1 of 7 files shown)


src/modules/InstShowInfo.rb
```diff
@@ -13,10 +14,24 @@ def main
...
+      @shown_info_files = []
     end
 
-    # @param [String] info_file path to be shown
-    def show_info_txt(info_file)
+    # @param file_path [String] path to be shown
+    def show_info_txt(file_path)
+      info_file = Y2Packager::InfoFile.read(file_path)
+      if info_file.nil?
+        Builtins.y2milestone("No %1", info_file)
```
dgdavid: :thinking: this will log "No nil", isn't it?

Probably a _leftover_. Maybe you wanted use _file_path_ instead

imobachgs: Yes, a leftover.

PR #491, opened 2019-12-05
Overwritable medium type

jreidinger:


src/lib/y2packager/medium_type.rb
```diff
@@ -33,6 +33,11 @@ def type
         @type ||= detect_medium_type
       end
 
+      # Allows to overwrite detected medium type. Useful e.g. when upgrade of
+      # registered system with Full medium should act like Online medium.
+      # @return [Symbol] possible values are `:online`, `:offline` and `:standard`
+      attr_writer :type
```
lslezak: Um, overriding the detected medium type is a quite dangerous operation, we should not do it silently.
I'd prefer some logging here.

And we should probably also validate the input value, to make sure it is one of the allowed symbols.

src/lib/y2packager/medium_type.rb
```diff
@@ -33,6 +33,22 @@ def type
...
+      POSSIBLE_TYPES = [:online, :offline, :standard].freeze
+
+      # Allows to overwrite detected medium type. Useful e.g. when upgrade of
+      # registered system with Full medium should act like Online medium.
+      # @param type [Symbol] possible values are `:online`, `:offline` and `:standard`
+      def type=(type)
+        log.info "Overwritting medium to #{type}"
+
+        if !POSSIBLE_TYPES.include?(type)
+          raise ArgumentError, "Not allows MediumType #{type.inspect}"
```
lslezak:
```suggestion
          raise ArgumentError, "Not allowed MediumType #{type.inspect}"
```

PR #492, opened 2019-12-05
Speed up calculation of libzypp product renames

imobachgs:
This PR solves [bsc#1157926](https://bugzilla.suse.com/show_bug.cgi?id=1157926). The real problem is that it takes too much time to calculate the product renames from `libzypp`. So this PRs adds some code to cache these renames.

On the other hand,  it avoids trying to find the product rename when the `product_package` property is missing, avoiding a lot of failing calls to `Pkg.ResolvableDependencies`.

In a test I did in a VM, the patch reduced the time to show the analysis result from 45 seconds to just 9 seconds. We can even improve this time by moving the logic to invalidate the cached data outside of the `AddOnProduct` module, avoiding asking libzypp for changes. However, that approach looks wrong to me.

Trello: https://trello.com/c/ZTskvrjr/

analyzing-your-system

## Speed comparison

<details>
<summary>Migration proposal (from 45s to 9s)</summary>

Old code: ~45 seconds
load-migration-products

New code: ~9 seconds
faster-load-migration-products
</details>

<details>
<summary>Migration target selection (from 12s to 4s)</summary>

Old code
load-migration-products

New code
faster-load-migration-products
</details>

### DIFF
--- a/src/lib/y2packager/repository.rb
+++ b/src/lib/y2packager/repository.rb
@@ -58,12 +58,13 @@ module Y2Packager
     class << self
       # Return all registered repositories
       #
+      # @param enabled_only [Boolean] Returns only enabled repositories
       # @return [Array<Repository>] Array containing all repositories
       #
       # @see Yast::Pkg.SourceGetCurrent
       # @see Y2Packager::Repository.find
-      def all
-        Yast::Pkg.SourceGetCurrent(false).map do |repo_id|
+      def all(enabled_only: false)
+        Yast::Pkg.SourceGetCurrent(enabled_only).map do |repo_id|
           find(repo_id)
         end
       end
--- a/src/lib/y2packager/repository.rb
+++ b/src/lib/y2packager/repository.rb
@@ -46,6 +46,8 @@ module Y2Packager
     attr_reader :name
     # @return [URI::Generic] Repository URL
     attr_reader :url
+    # @return [String] Product directory
+    attr_reader :product_dir
 
     attr_writer :enabled
     private :enabled=
@@ -91,7 +93,7 @@ module Y2Packager
         raise NotFound if repo_data.nil?
         new(repo_id: repo_id, enabled: repo_data["enabled"],
           name: repo_data["name"], autorefresh: repo_data["autorefresh"],
-          url: URI(repo_data["url"]))
+          url: URI(repo_data["url"]), product_dir: repo_data["product_dir"])
       end
 
       # Add a repository
@@ -101,12 +103,15 @@ module Y2Packager
       # @param autorefresh [Boolean]      Is auto-refresh enabled for this repository?
       # @param url         [URI::Generic] Repository URL
       # @return [Y2Packager::Repository,nil] New repository or nil if creation failed
-      def create(name:, url:, enabled: true, autorefresh: true)
+      def create(name:, url:, product_dir: "", enabled: true, autorefresh: true)
         repo_id = Yast::Pkg.RepositoryAdd(
           "name" => name, "base_urls" => [url], "enabled" => enabled, "autorefresh" => autorefresh
         )
         return nil unless repo_id
-        new(repo_id: repo_id, name: name, url: URI(url), enabled: enabled, autorefresh: autorefresh)
+        new(
+          repo_id: repo_id, name: name, url: URI(url), enabled: enabled,
+          autorefresh: autorefresh, product_dir: product_dir
+        )
       end
     end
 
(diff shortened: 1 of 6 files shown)


src/modules/AddOnProduct.rb
```diff
@@ -171,6 +171,8 @@ def main
       # E.g.: product:sle-module-basesystem-15-0.x86_64 has buddy
       # sle-module-basesystem-release-15-91.2.x86_64
       @selected_installation_products = [] # e.g.:  ["sle-module-basesystem"]
+
+      @libzypp_product_renames = nil
```
lslezak: Please a comment why we need it and what type it should be.

imobachgs:
> Maybe changing

[..]

> could speed it up a bit more as this would not fetch the unused attributes (like RPM size, path on the medium, package license, etc...) from libzypp.

Yes, perhaps it could help. Unfortunately, the new resolvables API is not available in SLE 15 SP1.

src/lib/y2packager/repository.rb
```diff
@@ -117,12 +123,14 @@ def create(name:, url:, enabled: true, autorefresh: true)
     # @param enabled     [Boolean]      Is the repository enabled?
     # @param autorefresh [Boolean]      Is auto-refresh enabled for this repository?
     # @param url         [URI::Generic] Repository URL
-    def initialize(repo_id:, name:, enabled:, autorefresh:, url:)
+    # @param product_dir [String]       Product directory
+    def initialize(repo_id:, name:, enabled:, autorefresh:, url:, product_dir: "/")
```
lslezak: Um, the `create` method uses an empty string as the default for `product_dir`, any reason for this inconsistency?

src/modules/AddOnProduct.rb
```diff
@@ -2198,6 +2205,21 @@ def y2update_path(src_id)
...
     end
+
+    # Determines whether the repositories libzypp information might have change
+    #
+    # This method returns true when the list of enabled repositories has changed.
+    # Disabled repositories are not even considered.
+    #
+    # @return [Boolean]
+    def libzypp_repos_changed?
+      repos = Y2Packager::Repository.all(enabled_only: true).sort_by(&:repo_id).map do |repo|
```
lslezak: I'd avoid sorting by `repo_id`, that's volatile and will change if you remove a repository and then add the same again. I'd rather sort the final URLs.

But there is a different problem, it's more complicated by using the DVD media. You might have two repositories with `dvd:///` URL but containing a completely different product/packages. So if you add a `dvd:///` repository, remove it and add it again but with a different medium in the drive you won't be able to detect the change if both use the same `/` product_dir... :worried: 

I do not know how to properly solve this, we would need to somehow digest the content of the medium (e.g. create a SHA sum of all `"<name>-<version>-<arch>"` packages on the medium or something like that and use that instead of the URL)...

imobachgs: Finally, I've decided to add the `repo_id` to the set of values to check so we are able to detect the `dvd:///` problem you have noticed.

test/addon_product_test.rb
```diff
@@ -63,6 +74,67 @@
...
+
+        it "does not ask libzypp again" do
+          expect(Yast::Pkg).to_not receive(:ResolvableDependencies)
+          subject.renamed?("old_name", new_product["name"])
+        end
+
+        context "but a new repo has been added" do
+          let(:repo1) do
+            instance_double(
+              Y2Packager::Repository, repo_id: 1, url: URI("dvd:///sr0"), product_dir: "/p1"
```
lslezak:
```suggestion
              Y2Packager::Repository, repo_id: 1, url: URI("dvd:///?devices=/dev/sr0"), product_dir: "/p1"
```

test/addon_product_test.rb
```diff
@@ -63,6 +74,67 @@
...
+
+          it "asks libzypp again" do
+            expect(Yast::Pkg).to receive(:ResolvableDependencies).and_return([])
+            subject.renamed?("old_name", new_product["name"])
+          end
+        end
+
+        context "but the url for a given repo has changed" do
+          before do
+            allow(repo0).to receive(:url).and_return(URI("dvd:///sr2"))
```
lslezak:
```suggestion
            allow(repo0).to receive(:url).and_return(URI("dvd:///?devices=/dev/sr2"))
```

test/addon_product_test.rb
```diff
@@ -31,6 +32,16 @@
...
 
     context "when according to libzypp a product is renamed" do
+      before do
+        subject.main
+        allow(Y2Packager::Repository).to receive(:all).and_return([repo0])
+      end
+
+      let(:repo0) do
+        instance_double(
+          Y2Packager::Repository, repo_id: 0, url: URI("dvd:///sr0"), product_dir: "/p0"
```
lslezak: I'd use some valid URL here, this might be confusing if somebody would copy&paste it to a real system...

```suggestion
          Y2Packager::Repository, repo_id: 0, url: URI("dvd:///?devices=/dev/sr0"), product_dir: "/p0"
```

PR #495, opened 2019-12-20
Config crash

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1073208

crash is caused by returning -1 on which Builtins.size raises exception.

### DIFF
--- a/src/modules/SourceManager.rb
+++ b/src/modules/SourceManager.rb
@@ -272,7 +272,9 @@ module Yast
 
       UI.CloseDialog if !Mode.commandline
 
-      if @newSources.nil? || @newSources == -1 || @newSources.size.zero?
+      # Pkg.SourceScan can return Array with repos or -1 if failed. So be more
+      # paranoid here and accept only non empty list as success result.
+      if !@newSources.is_a?(::Array) || @newSources.empty?
         message1 = Builtins.sformat(
           _("Unable to create repository\nfrom URL '%1'."),
           URL.HidePassword(url)
--- a/src/modules/SourceManager.rb
+++ b/src/modules/SourceManager.rb
@@ -272,7 +272,7 @@ module Yast
 
       UI.CloseDialog if !Mode.commandline
 
-      if Builtins.size(@newSources).zero?
+      if @newSources.nil? || @newSources == -1 || @newSources.size.zero?
         message1 = Builtins.sformat(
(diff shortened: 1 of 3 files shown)


src/modules/SourceManager.rb
```diff
@@ -272,7 +272,7 @@ def createSource(url)
 
       UI.CloseDialog if !Mode.commandline
 
-      if Builtins.size(@newSources).zero?
+      if @newSources.nil? || @newSources == -1 || @newSources.size.zero?
```
ancorgs: So `@newSources` can be nil, or -1, or a string... that's pretty weird already, so I would like to see more defensive code here (and a comment). This could still explode if `@newSources` would be, for example, -2. And looking at the history, that wouldn't surprise me.

jreidinger: well, I check pkg bindings about possible values and also plan to document it. I think in general this pkg-bindings API sucks, as it looks like in past it was list and empty list mean problem with source, but it was changed. So I plan second PR that will make it clear in Pkg method documentation.

ancorgs: And that's exactly why I don't trust the result of that call to `SourceScan` to stay consistent in the future. Of course, I trust you checked and the only possible values RIGHT NOW are those. But that changed already in the past and could change in the future.

So I would like to see a robust condition checking the result has the exact form we expect, raising the pop-up if not. Something like
`if !@newSources.kind_of?(::String) || @newSources.size.zero?`
or maybe
`if !@newSources.respond_to?(:size) || @newSources.size.zero?`

Just in case tomorrow somebody decides that some kind of errors should be reported as -2.

jreidinger: ok, in fact it expects List. See linked pr where I try to clarify it.

src/modules/SourceManager.rb
```diff
@@ -272,7 +272,9 @@ def createSource(url)
 
       UI.CloseDialog if !Mode.commandline
 
-      if Builtins.size(@newSources).zero?
+      # Pkg.SourceScan can return Array with repos or -1 if failed. So be more
+      # paranoid here and accept only non empty list as success result.
+      if !@newSources.is_a?(::Array) || @newSources.size.zero?
```
ancorgs: NP: `@newSources.empty?`

PR #496, opened 2019-12-27
Do not fail because %F "argument"

dgdavid:
## Problem

When the [Packager.desktop](https://github.com/yast/yast-packager/blob/master/desktop/org.opensuse.yast.Packager.desktop) file is _invoked_ without a list of files, YaST Software Manager fails with _"The following packages has not been found in the medium: %F"_ message.

* https://bugzilla.suse.com/show_bug.cgi?id=1159419

## Solution

Avoid to include the `%F` argument since it means that none file was given. Otherwise, it will be replaced by the list of files (see https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html for more information).

## Tests

Tested manually:

* To run the YaST Software Manager directly form the XFCE Settings Manager.
* To run the YaST Software Manger via right click -> "Open with YaST Software" in an _rpm_ file.

## Screenshots

<details>
<summary>Click to show/hide</summary>

---

<p align="center"><em>Before changes</em></p>

YaST Software Manager BEFORE changes

---

<p align="center"><em>After changes</em></p>

YaST Software Manager BEFORE changes

</details>

### DIFF
--- a/src/bin/sw_single_wrapper
+++ b/src/bin/sw_single_wrapper
@@ -8,8 +8,11 @@
 ARGS=""
 for ARG in "$@"
 do
-    QUOTED_ARG=$(printf %q "$ARG")
-    ARGS="$ARGS $QUOTED_ARG"
+    if [ "$ARG" != "%F" ]
+    then
+        QUOTED_ARG=$(printf %q "$ARG")
+        ARGS="$ARGS $QUOTED_ARG"
+    fi
 done
 
 xdg-su -c "/sbin/yast2 sw_single $ARGS"
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Dec 27 11:13:55 UTC 2019 - David Diaz <dgonzalez@suse.com>
+
+- Do not fail when Packager.desktop file is invoked without a list
+  of files (bsc#1159419).
+- 4.2.41
+
+-------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


src/bin/sw_single_wrapper
```diff
@@ -8,8 +8,11 @@
 ARGS=""
 for ARG in "$@"
 do
-    QUOTED_ARG=$(printf %q "$ARG")
-    ARGS="$ARGS $QUOTED_ARG"
+    if [ "$ARG" != "%F" ]
```
jreidinger: so do you test it and it really pass here "%F" ? that looks strange. How yast can know what is %F? It should be always expanded to list of arguments, not? If XFCE4 does not expand %F, then it is bug in XFCE and not in YaST.

dgdavid: > so do you test it and it really pass here "%F" ?

Yes, I did.

If necessary, I could upload _kind of logs_ here.

> If XFCE4 does not expand %F, then it is bug in XFCE and not in YaST.

Taking into account that the previous change in the YaST file was 2 years ago, a bug in XFCE sound reasonable to me.

Let's check it :)

jreidinger: what is the most importatnt it is violation of freedesktop specification. So if you can upload logs to bug report, I can redirect it to xfce maintainers.

dgdavid: Ok, I'll do

dgdavid:
Closing it because [it's XFCE not expanding %F](#discussion_r361645940). 

See also https://bugzilla.suse.com/show_bug.cgi?id=1159419#c10

PR #499, opened 2020-01-07
No medium1

jreidinger:
trello: https://trello.com/c/BzcjnWr2/1550-5-sles15-sp2-p2-1158498-build1011-media-name-on-installation-screen
bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1158498

## how it looks

TW before:
old

TW now (including new hourglass change):
Screenshot_2020-01-10_15-20-04

SLE before:
old2

SLE now:
Screenshot_2020-01-10_15-37-06

and new ncurses:
Screenshot_2020-01-10_15-47-14


src/modules/PackageSlideShow.rb
```diff
@@ -118,72 +114,46 @@ def GetPackageSummary
...
-          else
-            sum = Ops.add(sum, item)
-          end
-        end
-      end
-
-      sum = Ops.multiply(Ops.unary_minus(overflow), max_cutoff) if Ops.greater_than(overflow, 0)
-
-      sum
+      sizes.flatten.reduce(0) { |a, e| (e < 0) ? a : a + e }
```
schubi2: A comment would be nice how this is working :-)

src/modules/PackageSlideShow.rb
```diff
@@ -497,24 +425,15 @@ def FindNextMedia
...
-                "Next media: src: %1 CD: %2",
-                @next_src_no,
-                @next_cd_no
-              )
-            end
-            return
-          else
-            @next_cd_no = Ops.add(@next_cd_no, 1)
-          end
+          return if Ops.greater_than(Ops.get(remaining_sizes, @next_cd_no, 0), 0)
```
schubi2: Could we please get rid of Ops.gr.... ?

src/modules/PackageSlideShow.rb
```diff
@@ -497,24 +425,15 @@ def FindNextMedia
...
+
+          @next_cd_no += 1
         end
 
-        @next_src_no = Ops.add(@next_src_no, 1)
+        @next_src_no += 1
       end
 
-      Builtins.y2milestone("No next media - all done") if @debug
+      Builtins.y2milestone("No next media - all done")
```
schubi2: log.info would be nice

src/modules/PackageSlideShow.rb
```diff
@@ -873,70 +738,43 @@ def CdStatisticsTableItems
...
-              Builtins.sformat("src(%1)", src_no),
-              Ops.get(@inst_src_names, src_no, ""),
-              "",
-              "",
-              ""
+        next if ListSum(inst_src) < 1
+
+        cd_no = 0
+
+        Builtins.foreach(inst_src) do |src_remaining|
```
schubi2: Ruby style instead of Builtins.... ? :-)

src/modules/PackageSlideShow.rb
```diff
@@ -873,70 +738,43 @@ def CdStatisticsTableItems
...
+          if src_remaining > 0 ||
+              (src_no + 1) == @current_src_no &&
+                  (cd_no + 1) == @current_cd_no # suppress current CD
+            caption = @inst_src_names[src_no] || _("Unknown Source")
+            # add "Medium 1" only if more cds available (bsc#1158498)
+            caption += @media_type + (cd_no + 1).to_s unless @last_cd
+            rem_size = FormatRemainingSize(src_remaining) # column #1
+            rem_count = FormatRemainingCount(
+              Ops.get(@remaining_pkg_count_per_cd_per_src, [src_no, cd_no], 0)
```
schubi2: Ops.get....

src/modules/PackageSlideShow.rb
```diff
@@ -1008,51 +842,54 @@ def DisplayGlobalProgress
...
-              # progress message (part2)
-              Builtins.sformat(
-                _(" (Downloaded %1 of %2 packages)"),
-                @total_count_downloaded,
-                @total_count_to_download
-              )
+      @total_downloaded += @current_provide_size
+
+      @total_count_downloaded += 1
+      Builtins.y2milestone(
```
schubi2: log.info would be better

src/modules/PackageSlideShow.rb
```diff
@@ -1008,51 +842,54 @@ def DisplayGlobalProgress
...
+
+      d_mode = Ops.get_symbol(Pkg.CommitPolicy, "download_mode", :default)
+
+      if d_mode == :download_in_advance ||
+          d_mode == :default && Mode.normal &&
+              !Installation.dirinstall_installing_into_dir
+        # display download progress in DownloadInAdvance mode
+        # translations: progress message (part1)
+        SlideShow.SetGlobalProgressLabel(
+          Ops.add(
```
schubi2: Ops.add should be obsolete

src/modules/PackageSlideShow.rb
```diff
@@ -1008,51 +842,54 @@ def DisplayGlobalProgress
...
+            _("Downloading Packages..."),
+            # progress message (part2)
+            Builtins.sformat(
```
schubi2: I think we have "format" for that now.

src/modules/PackageSlideShow.rb
jreidinger: yeah, I should also improve method documentation that it sum only positive values.
Do you think that this code will be more readable? `sizes.flatten.select(:positive?).reduce(0, :+)`

src/modules/PackageSlideShow.rb
jreidinger: yes, will do.

src/modules/PackageSlideShow.rb
jreidinger: :+1:

src/modules/PackageSlideShow.rb
jreidinger: as it should be guranteed that inst_src is non-nil, I will change it.

src/modules/PackageSlideShow.rb
jreidinger: well, here I keep it intentionally as I do not want to break translation.

src/modules/PackageSlideShow.rb
schubi2: nice :-)

src/modules/PackageSlideShow.rb
schubi2: Well it is for master. So I am assuming changes anyway.

src/modules/PackageSlideShow.rb
```diff
@@ -8,17 +8,18 @@ class PackageSlideShowClass < Module
...
     # Column index for refreshing statistics: remaining size
     SIZE_COLUMN_POSITION = 1
     # Column index for refreshing statistics: remaining number of packages
     PKG_COUNT_COLUMN_POSITION = 2
     # Column index for refreshing statistics: remaining time
     TIME_COLUMN_POSITION = 3
+    # Table padding
+    ITEM_PREFIX = " " * 4
+    # hourglass unicode char
+    HOURGLASS = "\u231B".freeze
```
lslezak: Ruby uses by default UTF-8 for the sources, so we could use the literal character here. The advantage is that you can directly see how that looks like.

```suggestion
    HOURGLASS = "⌛".freeze
```

Or is there any reason for using the escape sequence?

src/modules/PackageSlideShow.rb
```diff
@@ -401,78 +355,52 @@ def InitPkgData(force)
...
       @last_cd = false
-      @unit_is_seconds = false # begin with package sizes
+      @unit_is_seconds = false # begin without showing time before it stabilize a bit
       @bytes_per_second = 1
 
       src_list = Pkg.PkgMediaNames
-      @inst_src_names = Builtins.maplist(src_list) do |src|
-        Ops.get_string(src, 0, "CD")
-      end
+      @inst_src_names = src_list.map { |src| src[0] || "CD" }
```
lslezak: NP: I usually prefer `.first`...

jreidinger: it later use [1] for different. So it is not first element, but index to some internal Pkg list. That is why I keep explicit 0.

src/modules/PackageSlideShow.rb
jreidinger: Well, reason for me is that it depends on used font how it will in the end look like. So I prefer have something that can be googled ( I know that even unicode char can be goodled ) and also something that is obvious that it is utf-8 char ( easy for this, but not so for e.g. … )

schubi2:
Ups, I have overseen the packaging stuff. Still LGTM :-)

PR #500, opened 2020-01-09
Do not refresh package installation overview if the medium has changed.

schubi2:
**Problem**
Reading Release Notes during installation is impossible because the installation switch to the
"Details" tab when the medium has been changed.
**Solution**
Do not refresh package installation overview if the medium has been changed and the user has switched to the release notes tab.

### DIFF
--- a/src/modules/PackageSlideShow.rb
+++ b/src/modules/PackageSlideShow.rb
@@ -540,6 +540,9 @@ module Yast
       SlideShow.CheckForSlides
       FindNextMedia()
 
+      # do not rebuild if the user reads the release notes
+      return if SlideShow.user_switched_to_release_notes
+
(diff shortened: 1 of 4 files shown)


src/modules/PackageSlideShow.rb
```diff
@@ -540,6 +540,9 @@ def SetCurrentCdNo(src_no, cd_no)
       SlideShow.CheckForSlides
       FindNextMedia()
 
+      # do not rebuild if the user reads the release notes
```
jreidinger: Hi, in general I do not like whole this piece of code ( nothing against your change ). It is from my POV touching internals of SlideShow module. I think ideally whole this part that basically only call various parts of SlideShow should live in Slideshow itself. ( including above called CheckForSlides ). Method like Redraw would be good.

PR #501, opened 2020-01-17
 [SLE-9109] Packages online search

imobachgs:
Please, see the whole picture in https://github.com/yast/yast-registration/pull/467.

This PR adds support to handle the online search dialog from the packager.

### DIFF
--- a/src/clients/sw_single.rb
+++ b/src/clients/sw_single.rb
@@ -456,7 +456,7 @@ module Yast
         end
       end
       repo_management = Mode.normal if repo_management.nil?
-      online_search = Mode.normal && Yast::WFM.ClientExists("online_search")
+      online_search = Mode.normal && registered?
 
       ret = {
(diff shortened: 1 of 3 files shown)


src/clients/sw_single.rb
```diff
@@ -456,8 +456,9 @@ def GetPackagerOptions
         end
       end
       repo_management = Mode.normal if repo_management.nil?
+      online_search = Mode.normal && Yast::WFM.ClientExists("online_search")
```
lslezak: What if the system is not registered at all? How is that handled?

imobachgs: It will be registered before registering the addons (see https://github.com/yast/yast-registration/pull/467/files#diff-57bedceb5faafb5d70b14b3357c09464R63).

lslezak: Please check also how it behaves in openSUSE Leap 15.1, you might create a reg code at https://scc.suse.com/admin/products/1929

PR #506, opened 2020-02-18
Avoid crashing when changing the language and keyboard layout

dgdavid:
## Problem

After code cleaning performed at https://github.com/yast/yast-packager/pull/499, `yast2 language` is crashing in `FindNextMedia` when changing the language and keyboard layout.

* https://bugzilla.suse.com/show_bug.cgi?id=1161842
* https://trello.com/c/4mvMZhMP/1607-ostumbleweed-p5-1161842-y2-packager-yast2-language-switch-get-error

## Solution

Do not try to find next media when there are no _remaining\_sizes\_per\_cd\_per\_src_

:warning: originally, the code [were falling back to default values](https://github.com/yast/yast-packager/blob/a14c1fcc11a0ce8547747daa8144a9bafbc81eef/src/modules/PackageSlideShow.rb#L409-L443), producing the same result than the `break` line introduced in this commit. :warning:

## Tests

- Only tested manually

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Feb 18 15:47:45 UTC 2020 - David Diaz <dgonzalez@suse.com>
+
+- Avoid crashing when changing a language and keyboard layout
+  (bsc#1161842).
+- 4.2.50
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


dgdavid:
> Downloaded 17 of 0 packages

Also the progress bar is always at 100%

dgdavid:
Thanks for the review @lslezak 

Let's make a friendly ping to @jreidinger here, just to not forget about those other issues. Maybe he could help when he is back.

PR #510, opened 2020-02-28
Empty preselected modules

imobachgs:
Do not crash when the list of preselected modules is empty (the default for `GetFeature` is an empty string).


src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -264,7 +264,7 @@ def preselected_upgrade_products
       # @return [Array<Y2Packager::ProductLocation>] the products
       def preselected_installation_products
         default_modules = Yast::ProductFeatures.GetFeature("software", "default_modules")
-        return [] unless default_modules
+        return [] if default_modules == ""
```
lslezak: Maybe we should check also for `nil` or possibly other types with:

```suggestion
        return [] unless default_modules.is_a?(Array)
```

PR #511, opened 2020-02-28
bsc#1157780: improve addon selector

imobachgs:
As the discussion about https://github.com/yast/yast-registration/pull/483 is taking some time, we have decided to propose this alternative fix until the new widget is ready. Bear in mind that the new widget will bring other improvements, but this PR should at least fix the bug described in [bsc#1157780](https://bugzilla.suse.com/show_bug.cgi?id=1157780) (although it changes the behaviour). Let's consider this a *Plan B* :)

In a nutshell, the idea is to consider the first available product as the current one (although it is not selected) and display its information. See the screenshots below and read on to find further information.

<details>
<summary>Addon selection in Qt</summary>

addon-selection-qt
</details>

<details>
<summary>Addon selection in ncurses</summary>

addon-selection-ncurses
</details>

## The problem

The selector widget considers the first element as the current one. So if the user clicks in the first element to see the details, the widget will not propagate that event (in reality, the value has not changed).

### DIFF
--- a/src/lib/y2packager/dialogs/addon_selector.rb
+++ b/src/lib/y2packager/dialogs/addon_selector.rb
@@ -108,7 +108,7 @@ module Y2Packager
         res = super
         Yast::Wizard.EnableNextButton
         Yast::Wizard.EnableBackButton
-
+        Yast::UI.SetFocus(Id(:addon_repos))
         res
       end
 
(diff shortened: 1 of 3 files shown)


src/lib/y2packager/dialogs/addon_selector.rb
```diff
@@ -206,16 +206,17 @@ def continue_msg
...
       # extra help text
-      # @return [String] translated text
-      def description
-        # TRANSLATORS: inline help text displayed below the product selection widget
-        _("Select a product to see its description here. The dependent products " \
-          "are selected automatically.")
+      # @return [String] first product description
+      def initial_description
+        return "" if products.empty?
+        product_description(products.first)
```
dgdavid: NP: Although the code will work as it is, maybe we could do that in a more explicit way, asking to the MultiSelectionBox by the :CurrentItem and looking for the product to display its description.

Anyway, I have no strong opinion about it and this changes look good enough to me to solve the issue mentioned in the PR description.

jreidinger:
for me change looks fine.

src/lib/y2packager/dialogs/addon_selector.rb
imobachgs: The problem is that, at this point, the products list is not displayed yet so you will get a `nil` value from `QueryWidget`.

PR #513, opened 2020-03-05
Avoid installation abort

joseivanlopez:
## Problem

When the user is adding an add-on and the repository cannot be created, a popup dialog gives to the user a new opportunity for changing the URL. If the user rejects such popup, the installation aborts instead of continue without adding the repository.

VirtualBox_SLE-15-SP2_05_03_2020_13_26_24

## Solution

Fix return symbol when the user rejects.

## Testing

* Added unit test.
* Tested manually.


imobachgs:
LGTM

PR #522, opened 2020-05-07
Crash on empty status file

mchf:
Bnc1166661

## Problem

when the file system_packages_repos.yaml was empty or somehow malformed (contained e.g. only '---') on first line yast-packager crashed with internal error as described in the bug (or a similar one).

## Solution

check what returns yaml parser

manually tested

### DIFF
--- a/src/lib/y2packager/known_repositories.rb
+++ b/src/lib/y2packager/known_repositories.rb
@@ -71,6 +71,12 @@ module Y2Packager
       end
 
       status = YAML.load_file(status_file)
+
+      if !status
+        log.info("Status file #{status_file} is empty or corrupted")
+        return []
+      end
+
       # unify the file in case it was manually modified
       status.uniq!
       status.sort!
--- a/src/lib/y2packager/known_repositories.rb
+++ b/src/lib/y2packager/known_repositories.rb
@@ -70,10 +70,15 @@ module Y2Packager
         return []
       end
 
-      status = YAML.load_file(status_file)
+      begin
+        status = YAML.load_file(status_file)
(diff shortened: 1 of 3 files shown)


src/lib/y2packager/known_repositories.rb
```diff
@@ -71,6 +71,12 @@ def read_repositories
       end
 
       status = YAML.load_file(status_file)
+
+      if !status
+        log.info("Status file #{status_file} is empty or corrupted")
+        return []
+      end
```
jreidinger, line 78: sadly, this is not enough, as Psych uses exceptions:

try yourself:
```
irb(main):002:0> YAML.load_file("/dev/zero")
Psych::SyntaxError ((/dev/zero): control characters are not allowed at line 1 column 1)
```
Your fix works only for empty or comment only `---` content.

mchf: well, in this case status file is hard coded and it doesn't seem to be user edited. 

moreover my fix was intended for content processing - parser itself didn't crash (no exception from the parser but nil / FalseClass exception on the line bellow)

but ok, i'll improve slightly more.

src/lib/y2packager/known_repositories.rb
```diff
@@ -70,10 +70,15 @@ def read_repositories
         return []
       end
 
-      status = YAML.load_file(status_file)
+      begin
+        status = YAML.load_file(status_file)
+      rescue Psych::SyntaxError
+        log.error("Status file #{status_file} is corrupted")
```
jreidinger, line 76: Looks good, just please also log exception content, so we know how it is corrupted.

PR #524, opened 2020-06-04
Drop anyxml

jreidinger:
trello: https://trello.com/c/Oi7jpQi0/1883-drop-old-xml-libraries

unit test included.

### DIFF
--- a/src/modules/OneClickInstallStandard.rb
+++ b/src/modules/OneClickInstallStandard.rb
@@ -1,10 +1,14 @@
 require "yast"
 
+require "nokogiri"
+
 # Yast namespace
 module Yast
   # Module for parsing One Click Install Standard
   # http://en.opensuse.org/Standards/One_Click_Install
   class OneClickInstallStandardClass < Module
+    include Yast::Logger
+
     def main
       textdomain "packager"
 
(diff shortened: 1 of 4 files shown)


src/modules/OneClickInstallStandard.rb
```diff
@@ -42,85 +46,55 @@ def main
     #              ],
     #      ]
     def GetRepositoriesFromXML(filename)
-      ret = []
 
       if !FileUtils.Exists(filename)
```
lslezak, line 50: NP: I can see a small inconsistency here. The `FileUtils.Exists` honors the SCR target chroot while the later `File.read` not. But because this is used only in an installed system we do not need to care about possible chroot.

So I'd simply replace `FileUtils.Exists` with standard Ruby `File.exist?`.

jreidinger: yes, possible. So lets do it.

test/one_click_install_standard_test.rb
```diff
@@ -0,0 +1,103 @@
...
+require "tempfile"
+
+Yast.import "OneClickInstallStandard"
+
+# note partially covered also by test/lib/clients/inst_productsources_test.rb
+describe "Yast::OneClickInstallStandard" do
+  subject { Yast::OneClickInstallStandard }
+
+  let(:file) do
+    File.expand_path("data/_openSUSE_Leap_15.0_Default.xml", __dir__)
```
mvidner, line 14: BTW did you know that Trang can infer a schema for a document?

For [that file](https://github.com/yast/yast-packager/blob/master/test/data/_openSUSE_Leap_15.0_Default.xml), here's what `trang _openSUSE_Leap_15.0_Default.xml _openSUSE_Leap_15.0_Default.rnc` produces, which is useful for sanity checking of the parsing code.

```rnc
namespace os = "http://opensuse.org/Standards/One_Click_Install"

start =
  element os:metapackage {
    element os:group {
      attribute distversion { text },
      element os:repositories {
        element os:repository {
          attribute alias { xsd:NCName }?,
          attribute format { xsd:NCName },
          attribute recommended { xsd:boolean },
          element os:name {
            attribute xml:lang { xsd:NCName }?,
            text
          }+,
          element os:summary {
            attribute xml:lang { xsd:NCName }?,
            text
          }+,
          element os:description {
            attribute xml:lang { xsd:NCName }?,
            text
          }+,
          element os:url { xsd:anyURI }
        }+
      }
    }
  }
```

jreidinger: interesting, we can use it now in installer to verify that xml is still same as we expect.

test/lib/clients/inst_productsources_test.rb
```diff
@@ -84,7 +85,10 @@ def list_of_repos
     end
 
     it "returns false if file is empty" do
-      expect(client.ParseListOfSources("/dev/zero", url)).to eq false
```
mvidner, line 87: Just pass "/dev/null", it is standard.

jreidinger: dev null is not for reading, only for writting, not?

mvidner: also for reading, https://man7.org/linux/man-pages/man4/null.4.html

PR #526, opened 2020-07-10
Improved medium type detection

lslezak:
## Introduction

Since SLE15-SP2 YaST needs to support Online, Full and standard (openSUSE Leap) installation media. See more details in the [media types documentation](https://github.com/yast/yast-installation/wiki/Installation-Media-Types).

The medium type is determined from the place where `install=` boot parameter points, it does not matter which medium was booted. You can boot the Online medium, but point `install=` to a repository containing the Full installation source.

## The Problem

See https://bugzilla.suse.com/show_bug.cgi?id=1173336

It turned out that a common use case is to mirror the repositories from SCC using SMT and then use a small installation medium for installation. You should not need to download that 10GB Full medium when you already mirrored all repositories from SCC, that makes sense.

So the use case is to mirror all repositories from SCC and point `install=` to the SMT repository with the SLES product.

Unfortunately it turned out that SMT does not mirror the `/media.1/products` file from SCC. This file is actually not needed for normal operations when accessing the repositories (installing packages). However, YaST uses that file for distinguishing between the Full and Online media.

It the file was missing YaST considered it as the Online installation medium. But the Online installation medium requires registration because it does not contain any packages to install. So the result was that YaST displayed an error popup that registration is mandatory although it was not true, the packages were available via the SMT repositories.

## The Fix

The fix is relatively simple, if YaST does not find any products (because the `/media.1/products` file is missing) it still tries to download the repository metadata and checks which products/packages are available there.

## TODO

- [ ] Report a bug for SMT, it should mirror this file as well...

### DIFF
--- a/src/lib/y2packager/repomd_downloader.rb
+++ b/src/lib/y2packager/repomd_downloader.rb
@@ -61,12 +61,18 @@ module Y2Packager
     # @return [Array<String>] List of paths pointing to the downloaded primary.xml.gz files,
     #   returns an empty list if the URL or the repository is not valid.
     #
-    def primary_xmls
-      return [] if product_repos.empty?
+    def primary_xmls(force = false)
+      repos = product_repos
+
+      if repos.empty?
+        return [] unless force
+
+        repos = { "" => "/" }
+      end
 
       # add a temporary repository for downloading the files via libzypp
       src = Yast::Pkg.RepositoryAdd("base_urls" => [url])
-      product_repos.map do |(_name, dir)|
+      repos.map do |(_name, dir)|
         # download the repository index file (repomd.xml)
         repomd_file = Yast::Pkg.SourceProvideFile(src, 1, File.join(dir, "repodata/repomd.xml"))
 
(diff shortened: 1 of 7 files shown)


src/lib/y2packager/repomd_downloader.rb
```diff
@@ -61,12 +64,20 @@ def product_repos
...
-      return [] if product_repos.empty?
+    def primary_xmls(force = false)
+      # first check if there are any products defined in /media.1/products
+      repos = product_repos
+
+      if repos.empty?
+        return [] unless force
+
+        # if `force` is true then scan the repository at the root anyway
+        repos = { "" => "/" }
```
jreidinger, line 75: well, I found quite strange to have empty string as key. Can we add some yast created name for repo? like product_root?

jreidinger: it is just NP

lslezak: If you check the code below you will see that the name is not used in the loop at all. We could place anything here, only the path is relevant for scanning the products. We could use some generic name like "Repository" but as it is ignored anyway I decided to use an empty string as the most simple name. :smiley:

lslezak: Ok, I have added a comment to make it clear.

PR #528, opened 2020-07-14
handle variable expansion in repository name

aschnell:
For https://bugzilla.suse.com/show_bug.cgi?id=1172477 and https://trello.com/c/QAAv9cMG/1895-3-osdistribution-p2-1172477-yast-replaces-releasever-by-the-current-release-in-repository-names-in-etc-zypp-reposd-repo.

The repository name can also include variables like $releasever or $arch. Handle and display raw_name where needed.

The extended yast2-pkg-bindings is needed.

### DIFF
--- a/src/include/packager/repositories_include.rb
+++ b/src/include/packager/repositories_include.rb
@@ -46,8 +46,8 @@ module Yast
     # @param plaindir [Boolean] true to use "PlainDir" format (no repository
     #   metadata present at the URL)
     # @param download [Boolean] whether to refresh the repository or not
-    # @param preffered_name [String] (optional) preferred repository name, use
-    #   empty string "" to generate the name
+    # @param preferred_name [String] (optional) preferred repository raw_name, use
+    #   empty string "" to generate the raw_name
     # @param force_alias [String] alias for the new repository, if a repository
     #   with this alias already exists then it is overwritten, use empty string ""
     #   to generate an unique alias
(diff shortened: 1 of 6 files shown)


src/include/packager/repositories_include.rb
```diff
@@ -46,8 +46,8 @@ def LicenseAccepted(id)
     # @param plaindir [Boolean] true to use "PlainDir" format (no repository
     #   metadata present at the URL)
     # @param download [Boolean] whether to refresh the repository or not
-    # @param preffered_name [String] (optional) preferred repository name, use
-    #   empty string "" to generate the name
+    # @param preferred_name [String] (optional) preferred repository raw_name, use
```
lslezak, line 49: Oh, thanks for fixing all typos and spelling issues...

src/lib/packager/clients/repositories.rb
```diff
@@ -362,6 +366,8 @@ def repoInfoRichText(name, category, info)
         "\">&nbsp;&nbsp;&nbsp;"
       )
 
+      name_string = name != raw_name ? _("Raw name: %s") % raw_name + "<BR>" : ""
```
lslezak, line 369: I think we should add a comment for translators so they know the context. Something like

```ruby
# TRANSLATORS: the raw name is the original name with unexpanded variables like "$releasever"
```

PR #530, opened 2020-07-15
handle variable expansion in repository name

aschnell:
For of https://github.com/yast/yast-packager/pull/528, https://github.com/yast/yast-packager/pull/529.

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Jul 14 10:41:22 CEST 2020 - aschnell@suse.com
+
+- Handle variable expansion in repository name (bsc#1172477)
+- 4.3.4
+
+-------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


package/yast2-packager.changes
```diff
@@ -2276,7 +2282,7 @@ Fri May 16 15:41:44 CEST 2014 - locilka@suse.com
 -------------------------------------------------------------------
 Fri May 16 12:31:52 UTC 2014 - jreidinger@suse.com
 
-- Fix preffering DVD during installation (bnc#865819,bnc#780617)
+- Fix preferring DVD during installation (bnc#865819,bnc#780617)
```
lslezak, line 2285: I'm not sure if OBS will accept this change, in the past there was a bot checking that the changes are only added at the top and the history was not changed. But I do not know if it is still used, maybe just try it and see...

aschnell: We will see if fixing typos is impossible.

PR #532, opened 2020-07-30
AY: Calling of post-partitioning scripts moved from dropped inst_autoimage …

schubi2:
…to inst_kickoff


src/clients/inst_kickoff.rb
```diff
@@ -77,6 +77,13 @@ def main
         # call it always, it handles installation mode inside
         WFM.CallFunction("inst_bootloader", WFM.Args)
         #  }
+
+        # calling postpartitioning scripts in AY installation mode
+        if Mode.autoinst
```
jreidinger, line 82: so this is not called during autoupgrade. What was previous state? was it called or not?

PR #533, opened 2020-09-02
Fixed name of the add-on repository (bsc#1175374)

lslezak:
## Summary

- Fix for https://bugzilla.suse.com/show_bug.cgi?id=1175374 - the add-on repository have different Product names when the YaST self-update is enabled.
    - Expected: "SUSE Linux Enterprise High Availability Extension 15 SP2"
    - Actual: "SLE-15-SP2-Product-HA-POOL-x86_64-Build99.9-Media1/"
- This was actually caused by this change https://github.com/yast/yast-pkg-bindings/pull/133

### Details

This is the problematic code snippet:

```ruby
sources = Yast::Pkg.SourceEditGet
sources.first["name"] = "new_name"
# does not change anything actually :-(
Yast::Pkg.SourceEditSet(sources)
```

The problem is that `Pkg.SourceEditGet` now returns two keys, `name` and `raw_name` and `Pkg.SourceEditSet` prefers the `raw_name` if it is present (see https://github.com/yast/yast-pkg-bindings/pull/133). So in this case the name is not changed because the original `raw_name` is preferred.

The fix is to set the preferred `raw_name` instead of `name`:
```ruby
sources.first["raw_name"] = "new_name"
```

Then it works as expected.

## Testing

Tested manually in three scenarios:

#### GA installer (self-update) disabled

This is the original behavior:

sle-15-sp2-addon-without_self_update

#### The Default Installation with Self-update

When running an SLE15-SP2 installation with enabled self-update you will see these repository names:

sle-15-sp2-addon-with_self_update

That's a regression from the GA behavior.

#### With the Fix

With this fix applied the previous names (as in GM) are displayed even with self-update enabled:

sle-15-sp2-addon-with_self_update_with_fix

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Sep  2 14:49:09 UTC 2020 - Ladislav Slezák <lslezak@suse.cz>
+
+- Fixed name of the add-on repository (bsc#1175374, related
+  to bsc#1172477)
+- 4.2.65
+
+-------------------------------------------------------------------
 Tue Jul 14 10:41:22 CEST 2020 - aschnell@suse.com
 
 - Handle variable expansion in repository name (bsc#1172477)
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.2.64
+Version:        4.2.65
 Release:        0
 Summary:        YaST2 - Package Library
 License:        GPL-2.0-or-later
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -1550,7 +1550,7 @@ module Yast
         sources_set = []
         Builtins.foreach(sources_got) do |one_source|
           if Ops.get_integer(one_source, "SrcId", -1) == src_id
-            Ops.set(one_source, "name", new_name)
+            Ops.set(one_source, "raw_name", new_name)
           end
           sources_set = Builtins.add(sources_set, one_source)
         end


mvidner:
https://yast-pkg-bindings.surge.sh/ does list `SourceEditGet`, `SourceEditSet`, but has no links to their docs :(

Do I understand it right? 
- `name` is like "SUSE Linux Enterprise High Availability Extension 15 SP2"
- `raw_name` is like "SLE-15-SP2-Product-HA-POOL-x86_64-Build99.9-Media1/"

lslezak:
- `raw_name` is "openSUSE Leap $releasever"
- `name` is "openSUSE Leap 15.2" (obviously when running in openSUSE Leap 15.2)

mvidner (approved):
OK, this is a bare minimum fix.

When I grep the sources for `SourceEditSet` or `name` or `raw_name` I get several hits and I cannot tell if we're missing something.

We definitely need to fix the missing SourceEditSet documentation (and the other missing builtins) at https://yast-pkg-bindings.surge.sh/ : the [text is in the sources](https://github.com/yast/yast-pkg-bindings/blob/03c80dcc8c11c7aa194dbe62a1686502ccb9d149/src/Source_Set.cc#L161) but it is not rendered. This bug would probably happen even with the docs present but this is an appropriate time to fix it. But it is in a different repo -> not this PR.

lslezak:
> When I grep the sources for `SourceEditSet` or `name` or `raw_name` I get several hits and I cannot tell if we're missing something.

I checked them all and found only one problematic place, fixed in https://github.com/yast/yast-add-on/pull/102.

lslezak:
Submitted in https://build.suse.de/request/show/226589

PR #537, opened 2020-10-08
bsc#1177485: do not install xorg-x11

imobachgs:
Fix [bsc#1177485](https://bugzilla.suse.com/show_bug.cgi?id=1177485). Do not install xorg-x11 when installation over VNC is enabled as it is not needed anymore.


imobachgs:
Superseded by https://github.com/yast/yast-packager/pull/538.

PR #538, opened 2020-10-12
VNC installation: Removed xorg-x11 package from the needed package li…

schubi2:
…st (bsc#1172459)

This package is not needed for an installation. Even AutoYaST has worked correctly :-)


wfeldt (approved):
lgtm

PR #539, opened 2020-10-26
Fix abort

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1178005

tested manually in installation


test/lib/dialogs/inst_product_license_test.rb
```diff
@@ -54,4 +54,12 @@
       dialog.contents
     end
   end
+
+  describe "#abort_handler" do
+    it "returns true if user confirm abort" do
```
teclator, line 59: np: would be nice to cover both cases, so it returns whether the user confirmed to abort or not....

teclator (approved):
LGTM

PR #544, opened 2021-01-15
Bring back commit 3344dfcd281e963b37

ancorgs:
## Problem

Commit 3344dfcd281e963b37 contained some code adjustments in the file `clients/software_proposal.rb` related to the introduction of storage-ng.

In parallel, commit 6a6b1764feabec3c10 moved that original file to a new location `lib/packager/clients/software_proposal.rb`.

As a consequence, when both branches were merged the changes of the former commit got lost and the file still contained the old code without the storage-ng adjustments.

## Solution

1) Checked why the problem has not been detected before.

The code was failing silently because `WMF.call` does not raise any error if the client doesn't exists. As a result, the method `#partitioning_changed?` was ALWAYS returning false.

2) Ensured `#partitioning_changed?` is the only affected code.

Done, that was the only modified file that has been moved afterwards, so everything else should be fine. Checked commit by commit in the original storage-ng branch.

3) Brought back the changes contained on that original commit.

Done, check next section for details.

## Testing

Manually tested. Verified that:

  - The first call to `#partitioning_changed?` always returns false. That's by design.
  - Subsequent calls return the expected result, i.e. true if the staging devicegraph has changed after the previous call to `#partitioning_changed?` and false otherwise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
 
     - name: Git Checkout
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
 
     # just for easier debugging...
     - name: Inspect Installed Packages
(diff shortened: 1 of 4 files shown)


joseivanlopez (approved):
LGTM

PR #549, opened 2021-02-15
Fix popup

jreidinger:
This pull request in yast-yast2 (https://github.com/yast/yast-yast2/pull/1122) changed the behavior during unit tests of `Yast::Report` and `Yast2::Popup`, making necessary to add some extra mocking to prevent YaST from trying to open windows during the test execution.

As far as we know, that affected the test-suites of: yast-storage-ng, yast-country, yast-firewall, yast-nfs-server, yast-nis-client, yast-pam, yast-security, yast-services-manager and yast-sysconfig.

This should fix the problem for this repository.

### DIFF
--- a/test/lib/widgets/product_license_confirmation_test.rb
+++ b/test/lib/widgets/product_license_confirmation_test.rb
@@ -125,7 +125,8 @@ describe Y2Packager::Widgets::ProductLicenseConfirmation do
       context "and license is still unconfirmed" do
         let(:license_confirmed?) { false }
 
-        it "returns false" do
+        it "returns false and Report it" do
+          expect(Yast::Report).to receive(:Message)
           expect(widget.validate).to eq(false)
         end
       end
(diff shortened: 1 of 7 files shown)


test/lib/widgets/product_license_confirmation_test.rb
```diff
@@ -125,7 +125,8 @@
       context "and license is still unconfirmed" do
         let(:license_confirmed?) { false }
 
-        it "returns false" do
+        it "returns false and Report it" do
```
ancorgs, line 128: NP: "reports the error and returns false"

ancorgs (approved):
I would like to see a bit more ad-hoc fixes targeting the exact individual problems of the tests, instead of throwing relatively general "allow(Report).to receive(:Error)" on whole contexts.

But taking into account the time pressure, LGTM.

test/lib/widgets/product_license_confirmation_test.rb
jreidinger: will do thanks.

jreidinger:
@ancorgs for other modules I try, but it problem is too complex or code too messy I use easier approach. So in general I agree with you.

PR #553, opened 2021-03-02
Add a drop-in replacement for dnf configuration cfa

hellcp:
This also should fix boo#1182849

### DIFF
--- a/src/lib/packager/clients/pkg_finish.rb
+++ b/src/lib/packager/clients/pkg_finish.rb
@@ -14,6 +14,7 @@
 require "installation/finish_client"
 require "y2packager/repository"
 require "packager/cfa/zypp_conf"
+require "packager/cfa/dnf_conf"
 
 Yast.import "InstURL"
 
@@ -32,6 +33,11 @@ module Yast
       "&& /bin/tar -czf '%<archive>s' '%<source>s'".freeze
     # Format of the timestamp to be used as repositories backup
     BACKUP_TIMESTAMP_FORMAT = "%Y%m%d-%H%M%S".freeze
+    # Map the CFAs to package managers
+    PKG_MAPPING = {
+      "dnf"     => Packager::CFA::DnfConf,
+      "libzypp" => Packager::CFA::ZyppConf
+    }.freeze
 
     # Constructor
     def initialize
@@ -100,7 +106,7 @@ module Yast
       # recommended packages, doc-packages,...
       # (needed for products like CASP)
       if ProductFeatures.GetBooleanFeature("software", "minimalistic_libzypp_config")
-        set_minimalistic_libzypp_conf
+        set_minimalistic_pkg_conf
       end
 
       # copy list of failed packages to installed system
@@ -284,15 +290,19 @@ module Yast
       Pkg.TargetInitialize(Installation.destdir)
     end
 
-    # Set libzypp configuration to install the minimal amount of packages
+    # Set package manager configuration to install the minimal amount of packages
     #
     # @see Yast::Packager::CFA::ZyppConf#set_minimalistic!
-    def set_minimalistic_libzypp_conf
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
-      config.load
-      config.set_minimalistic!
-      config.save
+    def set_minimalistic_pkg_conf
+      PKG_MAPPING.each do |pkg, cfa|
+        next unless File.file?(File.join(Installation.destdir, cfa::PATH))
+
+        log.info("Setting #{pkg} configuration as minimalistic")
+        config = cfa.new
+        config.load
+        config.set_minimalistic!
+        config.save
+      end
     end
   end
 end
--- a/test/pkg_finish_test.rb
+++ b/test/pkg_finish_test.rb
@@ -344,7 +344,7 @@ describe Yast::PkgFinishClient do
         let(:destdir) { tmpdir }
 
         before do
-          conf_file = tmpdir.join(Yast::Packager::CFA::ZyppConf::PATH[1..-1])
+          conf_file = File.join(tmpdir, Yast::Packager::CFA::ZyppConf::PATH)
           FileUtils.mkdir_p(File.dirname(conf_file))
           FileUtils.touch(conf_file)
         end
--- a/test/pkg_finish_test.rb
+++ b/test/pkg_finish_test.rb
@@ -341,6 +341,13 @@ describe Yast::PkgFinishClient do
 
       context "if libzypp's minimalistic configuration is enabled" do
         let(:minimalistic_libzypp_config) { true }
+        let(:destdir) { tmpdir }
+
+        before do
+          conf_file = tmpdir.join(Yast::Packager::CFA::ZyppConf::PATH[1..-1])
+          FileUtils.mkdir_p(File.dirname(conf_file))
+          FileUtils.touch(conf_file)
+        end
 
         it "sets libzypp configuration to be minimalistic" do
           expect(zypp_conf).to receive(:set_minimalistic!)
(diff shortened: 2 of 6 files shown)


Conan-Kudo (approved):

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -288,8 +289,15 @@ def sync_target_sources
     #
     # @see Yast::Packager::CFA::ZyppConf#set_minimalistic!
     def set_minimalistic_libzypp_conf
```
lslezak, line 291: As it is now more generic I'd rename the method. It's safe, it's a private method.

```suggestion
    def set_minimalistic_pkg_conf
```

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -288,8 +289,15 @@ def sync_target_sources
...
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
+      config = if File.file?("/etc/zypp/zypp.conf")
+        log.info("Setting libzypp configuration as minimalistic")
+        Packager::CFA::ZyppConf.new
+      elsif File.file?("/etc/dnf/dnf.conf")
+        log.info("Setting dnf configuration as minimalistic")
+        Packager::CFA::DnfConf.new
+      end
+      return if config.nil?
```
jreidinger, line 299: I would like to see this also in logs. So maybe something like
```
else
  log.info "No packager config found. Not setting minimalistic packager configuration."
  return
end
```

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -288,8 +289,15 @@ def sync_target_sources
     #
     # @see Yast::Packager::CFA::ZyppConf#set_minimalistic!
     def set_minimalistic_libzypp_conf
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
+      config = if File.file?("/etc/zypp/zypp.conf")
+        log.info("Setting libzypp configuration as minimalistic")
+        Packager::CFA::ZyppConf.new
+      elsif File.file?("/etc/dnf/dnf.conf")
```
lslezak, line 295: I do not know if it makes sense to have dnf and zypper installed in parallel, but what about adapting both if they are present?

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -288,8 +289,15 @@ def sync_target_sources
     #
     # @see Yast::Packager::CFA::ZyppConf#set_minimalistic!
     def set_minimalistic_libzypp_conf
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
+      config = if File.file?("/etc/zypp/zypp.conf")
```
lslezak, line 292: I think there is a missing chroot prefix, the real path during installation is `/mnt/etc/zypp/zypp.conf`.

So this should be 

```suggestion
      config = if File.file?(File.join(Installation.destdir, "/etc/zypp/zypp.conf"))
```

lslezak (commented):
Missing `/mnt/` prefix?

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -284,15 +285,24 @@ def sync_target_sources
...
     # @see Yast::Packager::CFA::ZyppConf#set_minimalistic!
-    def set_minimalistic_libzypp_conf
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
-      config.load
-      config.set_minimalistic!
-      config.save
+    def set_minimalistic_pkg_conf
+      ["dnf", "libzypp"].each do |pkg|
+        file = "/etc/dnf/dnf.conf" if pkg == "dnf"
```
lslezak, line 293: Use the constants here (`Packager::CFA::DnfConf::PATH`) the path should be ideally defined in a single place to minimize possible changes.

jreidinger: I would go even one step further and define constant mapping
```
PKG_MAPPING = {
  "dnf" => "/etc/dnf/dnf.conf",
  "libzypp" => "/etc/zypp/zypp.conf"
}
```

and then PKG_MAPPING.keys.each do |pkg| and so on.

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -284,15 +285,24 @@ def sync_target_sources
...
     #
     # @see Yast::Packager::CFA::ZyppConf#set_minimalistic!
-    def set_minimalistic_libzypp_conf
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
-      config.load
-      config.set_minimalistic!
-      config.save
+    def set_minimalistic_pkg_conf
+      ["dnf", "libzypp"].each do |pkg|
```
lslezak, line 292: I do not like these artificial names , what about iterating over the class names instead?

```ruby
[ Packager::CFA::ZyppConf, Packager::CFA::DnfConf].each do |conf_class|
  next unless File.file?(File.join(Installation.destdir, conf_class.const_get(:PATH))
  config = conf_class.new
  config.load
  config.set_minimalistic!
  config.save
end
```

hellcp: Names are now only used for logging, that should work

lslezak: `conf_class::PATH` should also work (instead of that `conf_class.const_get`).

src/lib/packager/clients/pkg_finish.rb
```diff
@@ -284,15 +290,18 @@ def sync_target_sources
...
-    def set_minimalistic_libzypp_conf
-      log.info("Setting libzypp configuration as minimalistic")
-      config = Packager::CFA::ZyppConf.new
-      config.load
-      config.set_minimalistic!
-      config.save
+    def set_minimalistic_pkg_conf
+      PKG_MAPPING.each do |pkg, cfa|
+        next unless File.file?(File.join(Installation.destdir, cfa::PATH))
+        log.info("Setting #{pkg} configuration as minimalistic")
```
lslezak, line 299: Nitpicking: if use `#{cfa::PATH}` here then we do not need the `pkg` values at all and we could replace the Hash with an Array.... :grin:

test/pkg_finish_test.rb
```diff
@@ -341,6 +341,14 @@
 
       context "if libzypp's minimalistic configuration is enabled" do
         let(:minimalistic_libzypp_config) { true }
+        let(:destdir) { tmpdir }
+
+        before do
+          conf_file = Yast::Packager::CFA::ZyppConf::PATH[1..-1]
+          conf_path = conf_file.split("/")[0..-2].join("/")
```
lslezak, line 348: This looks too complicated and it just returns the "etc/zypp" path. What about simplifying it to:

```suggestion
          conf_path = File.dirname(Yast::Packager::CFA::ZyppConf::PATH)
```

`File.join` does not mind if there is an extra `/`...

hellcp:
I tested this on system with just dnf, and it installs just fine, and sets the correct config
Screenshot from 2021-03-04 23-32-21

test/pkg_finish_test.rb
```diff
@@ -341,6 +341,13 @@
 
       context "if libzypp's minimalistic configuration is enabled" do
         let(:minimalistic_libzypp_config) { true }
+        let(:destdir) { tmpdir }
+
+        before do
+          conf_file = tmpdir.join(Yast::Packager::CFA::ZyppConf::PATH[1..-1])
```
lslezak, line 347: You can use `File.join` to avoid that ugly `[1..-1]`:

```suggestion
          conf_file = File.join(tmpdir, Yast::Packager::CFA::ZyppConf::PATH)
```

That looks much better IMHO...

lslezak (approved):
LGTM

lslezak:
@hellcp thank you! :+1:

PR #559, opened 2021-03-11
Move the libzypp cache (bsc#1182928)

lslezak:
- Move the libzypp cache to the installed system to save some memory and avoid crashing because out of memory (bsc#1182928)
  - Create a symlink from the inst-sys to `/mnt` so the running libzypp can still find the files
- See related https://github.com/yast/yast-installation/pull/932
- 4.3.19

## Notes

For copying the cache we use the `Pkg.SourceCacheCopyTo()`, we already used it but at the very end of the installation. Just move it earlier.

This `Pkg` call also copies the zypp credentials so we can drop the old code.

## Testing

Tested manually in a Leap 15.3, SLE15-SP3 Full (DVD) and Online (SCC) installation.

### DIFF
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.3.18
+Version:        4.3.20
 Release:        0
 Summary:        YaST2 - Package Library
 License:        GPL-2.0-or-later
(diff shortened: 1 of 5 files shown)


package/yast2-packager.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.3.18
+Version:        4.3.20
```
mvidner, line 20: .19
bumped twice?

mvidner (changes requested):
code looks good but the packaging has glitches

mvidner (approved):

PR #560, opened 2021-03-18
Zypp cache symlink fix, removed "repositories" button in the console

lslezak:
- Fixed symlink issue in installed system
  - Create the symlink only in `Stage.initial`, when running in the inst-sys so it creates correct
    `/var/cache/zypp` -> `(/mnt)/var/cache/zypp` symlink
  - In installed system it would create  `/var/cache/zypp` -> `(/)/var/cache/zypp` symlink which creates a loop
  - https://bugzilla.suse.com/show_bug.cgi?id=1183683
- Removed "repositories" button in the installer console, it's too dangerous
  - https://bugzilla.suse.com/show_bug.cgi?id=1183687


src/clients/inst_kickoff.rb
```diff
@@ -46,15 +46,18 @@ def main
...
+        # copy the credential files (libzypp loads them from target)
+        # and the repository cache to the target system
+        Pkg.SourceCacheCopyTo(Installation.destdir)
+
+        # symlink the cache from inst-sys to save same space (RAM!)
+        cache_path = Pkg.ZConfig()["repo_cache_path"] || "/var/cache/zypp"
+        log.info("Zypp cache size: #{`du -h -s #{cache_path.shellescape}`}")
+        ::FileUtils.rm_rf(cache_path)
+        File.symlink(File.join(Installation.destdir, cache_path), cache_path)
+      end
```
dgdavid, line 60: NP: I'd move this kind of blocks to their own method, using a guard clause at the very top to avoid its execution if not running in the expected stage. You know, something like 

```ruby
# Copy the zypp cache to the target system (only when running in inst-sys!)
# 
# More documentation, if needed
# ...
#
def copy_zypp_cache
  return unless Stage.initial

  ...
end
```

For me is more readable and helps to document and understand the code :)

dgdavid (approved):
LGTM

dgdavid (approved):

PR #561, opened 2021-03-31
Delete zypp cache in the destination before overwriting it (boo#1183711)

Vogtinator:
When the copy happens, libzypp already has the cache file in the destination
open, but the copy from the local cache overwrites the opened file. This leads
to data corruption and ultimately crashes.

### DIFF
--- a/src/clients/inst_kickoff.rb
+++ b/src/clients/inst_kickoff.rb
@@ -136,13 +136,18 @@ module Yast
     def copy_zypp_cache
       return unless Stage.initial
 
+      cache_path = Pkg.ZConfig()["repo_cache_path"] || "/var/cache/zypp"
+      log.info("Zypp cache size: #{`du -h -s #{cache_path.shellescape}`}")
+
+      # Delete the destination cache for consistency and to avoid that the copy
+      # below modifies opened files in place, leading to corruption.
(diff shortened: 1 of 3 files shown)


src/clients/inst_kickoff.rb
```diff
@@ -136,13 +136,18 @@ def main
     def copy_zypp_cache
       return unless Stage.initial
 
+      cache_path = Pkg.ZConfig()["repo_cache_path"] || "/var/cache/zypp"
+      log.info("Zypp cache size: #{`du -h -s #{cache_path.shellescape}`}")
+
+      # Delete the destination cache for consistency and to avoid that the copy
+      # below modifies opened files in place, leading to corruption.
+      ::FileUtils.rm_rf(File.join(Installation.destdir, cache_path))
```
ancorgs, line 144: Shouldn't `SourceCacheCopyTo` take care of this?

Vogtinator: It could just pass `--remove-destination` to `cp`, yes.

If you prefer that instead and this PR is still useful despite the latest comments in the bug report, I'll change it

lslezak:
As Michael [commented](https://bugzilla.suse.com/show_bug.cgi?id=1183711#c20) we should not touch the cache at all. I have reverted the change in #562, this is not needed anymore.

PR #568, opened 2021-05-26
Use "armv7hl" packages on "armv7l" (bsc#1183795)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1183795#c16
- We need to handle the `armv7l` architecture specifically
- A [similar workaround](https://github.com/openSUSE/libzypp/blob/7f345ea4892fd02345e8de47c2a08ab5b174650b/zypp/ZConfig.cc#L145-L146) is applied in libzypp

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu May 27 07:43:27 UTC 2021 - Ladislav Slezák <lslezak@suse.cz>
+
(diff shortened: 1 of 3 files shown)


ggardet:
I tested it successfully on armv7 in qemu.

dgdavid (approved):
LGTM

lslezak:
Submitted manually in https://build.suse.de/request/show/241906

PR #575, opened 2021-07-27
Fixed evaluating module dependencies on the SLE Full medium 

lslezak:
## The Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1188633
  - Esp. see comments [c#10](https://bugzilla.suse.com/show_bug.cgi?id=1188633#c10) and [c#14](https://bugzilla.suse.com/show_bug.cgi?id=1188633#c14)
- After selecting the "Workstation Extension" the "Python2" module was automatically selected
- It turned out that there is a soft dependency in the WE product:
  `Recommends:       product(sle-module-python2) >= 15.4`
  But that should not select the module automatically, it's not a hard requirement, just recommended. 
- The Online medium uses only the hard dependencies (from SCC), the Full medium should behave similarly

## The Fix

- Ignore soft dependencies (`Recommends`) in the solver, use only the hard dependencies (`Requires`)

## Screenshots

Original behavior, the "Python2" module was automatically selected for the "Workstation Extension":
we_selection_orig

Fixed version, the "Python2" module is not autoselected anymore:
we_selection_fixed

## Testing

Tested manually in the latest Full medium build, see the screenshots above.

### DIFF
--- a/src/lib/y2packager/product_finder.rb
+++ b/src/lib/y2packager/product_finder.rb
@@ -184,6 +184,9 @@ module Y2Packager
       jobs = select_products(product_solvable, selected_base)
       solver = pool.Solver
 
+      # disable soft dependencies (Recommends), use only the hard dependencies (Requires)
+      solver.set_flag(Solv::Solver::SOLVER_FLAG_IGNORE_RECOMMENDED, 1)
+
       # run the solver to evaluate all dependencies
       problems = solver.solve(jobs)
 
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Tue Aug 17 15:18:07 UTC 2021 - Ladislav Slezák <lslezak@suse.cz>
+
+- Fixed evaluating module dependencies on the SLE Full medium
+  - Ignore soft dependencies (Recommends), use only hard
+    dependencies (Requires)
+  - Do not select "Python2" module for "Workstation Extensions"
+    (bsc#1188633)
+- 4.4.8
+
+-------------------------------------------------------------------
 Thu Aug 12 13:28:52 UTC 2021 - Ladislav Slezák <lslezak@suse.cz>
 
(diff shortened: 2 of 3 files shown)


mvidner:
I've commented in the bug, c8:

> OK, let me question the validity of this change. 
> I think that yast works as intended in preselecting the Python module. The user can unselect it without causing a dependency problem. If it shouldn't be preselected, just drop the Recommends dependency. Am I missing something?

src/lib/y2packager/product_finder.rb
```diff
@@ -184,6 +184,9 @@ def find_dependencies(product_solvable, selected_base)
       jobs = select_products(product_solvable, selected_base)
       solver = pool.Solver
 
+      # disable soft dependencies (Recommends), use only the hard dependencies (Requires)
+      solver.set_flag(Solv::Solver::SOLVER_FLAG_IGNORE_RECOMMENDED, 1)
```
mvidner, line 188: What is the lifetime of this flag setting?
Will it affect package resolving further on?
Is libzypp using the same solver instance as ruby-solv?

(maybe inconsequential as I want to cancel the whole PR)

lslezak: Libzypp uses a different solver instance, there is no side effect of this change.

lslezak:
> > I think that yast works as intended in preselecting the Python module. The user can unselect it without causing a dependency problem. If it shouldn't be preselected, just drop the Recommends dependency. Am I missing something?

Unfortunately it cannot be deselected, YaST treats the dependency as hard, see [c#10](https://bugzilla.suse.com/show_bug.cgi?id=1188633#c10).

It also [turned out](https://bugzilla.suse.com/show_bug.cgi?id=1188633#c14) that in Online installation the SCC reports only the hard dependencies so the Python module is not selected in Online installation.

We should unify the behavior, the dependencies on Full medium should ideally work exactly same as on Online medium.

shundhammer (approved):
LGTM

PR #576, opened 2021-08-12
Adjusted memory check for RPi (bsc#1139325)

lslezak:
- Related to https://bugzilla.suse.com/show_bug.cgi?id=1139325
- YaST displays a [warning on low memory systems](https://github.com/yast/yast-packager/blob/2c0c05736812df34f7bd4031e8dc0067d892881c/src/lib/y2packager/clients/inst_productsources.rb#L1684-L1690) to discourage using the online repositories (the repository metadata need quite a lot of RAM)
- But on Raspberry Pi the amount of memory exactly matches the limit so no warning is displayed (if it had *one byte less* memory the warning would be displayed)
- Update the check to display the warning in that case as well

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 12 13:28:52 UTC 2021 - Ladislav Slezák <lslezak@suse.cz>
+
+- Adjusted memory check to display a low memory warning on RPi
+  (bsc#1139325)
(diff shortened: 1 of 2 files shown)


dgdavid (approved):
LGTM

dgdavid:
@lslezak please check the [comment in OBS request](https://build.opensuse.org/request/show/911842#comment-1503273)

> Causes low memory warning in openQA, might need changes (and/or discussion)

PR #580, opened 2021-10-13
Add the product specification concept

imobachgs:
Trello: https://trello.com/c/oBnHQyrM/

## The problem

The `Y2Packager` module contains three classes to represent a product, depending on where they are read from. Initially, the idea of this PR was to unify them into the same `Y2Packager::Product` class. However, we found out that `ProductControlProduct`, `ProductLocation` and `Product` are not exactly the same.

We identified two concepts: the product *specification* (`ProductLocation` and `ProductControlProduct` belongs to that concept) and the *real product* (the libzypp incarnation of the product, represented by the `Product` class). Once you select a product coming from the control file or the full-medium, you need to do some work in advance (like registering or adding the repository) to be able to select the *real product*.

## The proposed solution

We have introduced a set of classes representing the *product specification*:

* `ProductSpec` is the base product.
* `RepoProductSpec` represents the product specification coming from the full medium (it adds some attributes like `dir`).
* `ControlProductSpec` represents the specification coming from the control file.
* `LibzyppProductSpec` represents... well, a product specification coming from libzypp.

All of them implement a `#select` method which implements the logic to select the product. This logic was included in the `complex_welcome` client initially. Probably, once we come up with a solution to https://trello.com/c/it2eG7WI/ it will be moved to a better place.

Additionally, a method `#to_product` converts the specification to a proper `Y2Packager::Product`.

These changes have a direct impact on `yast2-installation`. See this [add-product-spec branch](https://github.com/yast/yast-installation/compare/add-product-spec?expand=1) for the ongoing changes (WIP).

### DIFF
--- a/src/lib/y2packager/product_spec.rb
+++ b/src/lib/y2packager/product_spec.rb
@@ -76,6 +76,27 @@ module Y2Packager
       # Resets the products cache
       def reset
         @products = nil
+        @forced_base_product = nil
+      end
+
+      # Returns, if any, the base product which must be selected
+      #
+      # A base product can be forced to be selected through the `select_product`
+      # element in the software section of the control.xml file (bsc#1124590,
+      # bsc#1143943).
+      #
+      # @return [Y2Packager::Product, nil] the forced base product or nil when
+      # either, it wasn't selected or the selected wasn't found among the
+      # available ones.
+      def forced_base_product
+        return @forced_base_product if @forced_base_product
+
+        Yast.import "ProductFeatures"
+
+        forced_product_name = Yast::ProductFeatures.GetStringFeature("software", "select_product")
+        return if forced_product_name.to_s.empty?
+
+        @forced_base_product = base_products.find { |p| p.name == forced_product_name }
       end
     end
 
--- a/test/lib/product_spec_test.rb
+++ b/test/lib/product_spec_test.rb
@@ -67,4 +67,58 @@ describe Y2Packager::ProductSpec do
       end
     end
   end
+
+  describe ".forced_base_product" do
+    let(:select_product) { nil }
+
+    let(:opensuse) do
+      instance_double(Y2Packager::ProductSpec, name: "openSUSE")
+    end
+
+    let(:sle) do
+      instance_double(Y2Packager::ProductSpec, name: "SLE")
+    end
+
+    before do
+      described_class.reset
+
+      allow(described_class).to receive(:base_products).and_return([opensuse, sle])
+
+      allow(Yast::ProductFeatures).to receive(:GetStringFeature)
+        .with("software", "select_product")
+        .and_return(select_product)
+    end
+
+    context "when the control file is not forcing to select a base product selected" do
+      it "returns nil" do
+        expect(described_class.forced_base_product).to be_nil
+      end
+    end
+
+    context "when the control file is not forcing to select a base product selected" do
+      context "and the product is available" do
+        let(:select_product) { "openSUSE" }
+
+        it "returns the prodcut" do
+          expect(described_class.forced_base_product).to eq(opensuse)
+        end
+      end
+
+      context "but none available base product name match" do
+        let(:select_product) { "Whatever product" }
+
+        it "returns nil" do
+          expect(described_class.forced_base_product).to be_nil
+        end
+      end
+
+      context "but is empty" do
+        let(:select_product) { "" }
+
+        it "returns nil" do
+          expect(described_class.forced_base_product).to be_nil
+        end
+      end
+    end
+  end
 end


src/lib/y2packager/control_product_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2packager/product_spec"
+
+module Y2Packager
+  # Describes a product for installation that comes from the XML definition
```
jreidinger, line 23: I think important info is that it is not XML, but that it is from control.xml file.

src/lib/y2packager/product_spec_readers/control.rb
```diff
@@ -0,0 +1,80 @@
...
+    # Reads product specifications from the control file
+    class Control
+      include Yast::Logger
+
+      # map the Arch.architecture to the arch expected by SCC
+      REG_ARCH = {
+        "s390_32" => "s390",
+        "s390_64" => "s390x",
+        # ppc64le is the only supported PPC arch, we do not have to distinguish the BE/LE variants
+        "ppc64"   => "ppc64le"
```
jreidinger, line 38: I am a bit suspicious that it is the first time we need such mapping in YaST. So I expect it has to live also somewhere else.

src/lib/y2packager/repo_product_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      @depends_on = depends_on
+      @dir = dir
+      @media_name = media_name
+      @description = description
+    end
+
+    def select
+      super
+
+      # in offline installation add the repository with the selected base product
```
jreidinger, line 62: To be honest I do not like much this logic here in spec. It contains a lot of logic that is not related to Multi Repo. Why e.g. only thi spec initialize addon and workflow manager? It should be done for all products. Especially when there is reset method, which does not reset it.

src/lib/y2packager/repo_product_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      show_popup = true
+      base_url = Yast::InstURL.installInf2Url("")
+      log_url = Yast::URL.HidePassword(base_url)
+      Yast::Packages.Initialize_StageInitial(show_popup, base_url, log_url, dir)
```
jreidinger, line 66: just curious, but why always pass log_url here? Why not hide password inside method?

src/lib/y2packager/product_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+require "y2packager/product_license_mixin"
+
+module Y2Packager
+  # Describes a product that the user can select for installation
+  #
+  # The installer allows selecting different products for installation. The list of products can be
+  # read from different places (the libzypp database, the control.xml file, etc.). This class
+  # represents those products, read from different places.
+  #
+  # Bear in mind that, once selected, the product is mapped to a proper Y2Packager::Product class,
```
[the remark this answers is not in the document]
imobachgs: Yes, in that case, it is not possible. I will clarify.

src/lib/y2packager/product_spec_readers/control.rb
imobachgs: Yes, I will add a note to address it in a near future. It was taken [from here](https://github.com/yast/yast-yast2/blob/4f0560a55f0b4ece41ed50063fd7baf00cd8eedc/library/packages/src/lib/y2packager/product_control_product.rb#L35-L40).

src/lib/y2packager/product_spec_reader.rb
```diff
@@ -0,0 +1,71 @@
...
+    #
+    # @return [Y2Packager::ProductSpec] List of product specifications
+    def products
+      # products_from_control || products_from_offline || products_from_libzypp
+
+      case Y2Packager::MediumType.type
+      when :online
+        products_from_control
+      when :offline
+        products_from_offline
```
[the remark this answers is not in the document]
imobachgs: Yes, I agree. Whether the medium is online/offline/standard should not be that important at this point, but if we are dealing with multi-repos, single repo or control file based products.

src/lib/y2packager/repo_product_spec.rb
imobachgs: I expect to address this in https://trello.com/c/it2eG7WI/

src/lib/y2packager/repo_product_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      @dir = dir
+      @media_name = media_name
+      @description = description
+    end
+
+    def select
+      super
+
+      # in offline installation add the repository with the selected base product
+      show_popup = true
```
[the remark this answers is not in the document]
imobachgs: Same as https://github.com/yast/yast-packager/pull/580#pullrequestreview-779014276. I do not like it either, but I see it as a first step. Now all the related logic is at the same place and we will be able to find a better home for them :-)

PR #588, opened 2021-11-04
Adjusted low memory message (bsc#1139325)

lslezak:
- In #576 the memory check was adjusted so that the computer memory must be greater than the limit set, in the past it was enough if the size was the same as the limit.
- Adjust the message so it needs more memory than the limit as well.
- 4.4.11


joseivanlopez (approved):
LGTM

PR #592, opened 2021-11-19
Repository names

lslezak:
## The Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1191652
- The repositories added from the Full medium during installation have different names than the same repositories added in installed system.

repo_names_broken_sp3

```
# zypper lr
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias                             | Name                           | Enabled | GPG Check | Refresh
--+-----------------------------------+--------------------------------+---------+-----------+--------
1 | Basesystem-Module_15.3-0          | sle-module-basesystem          | No      | ----      | ----
2 | SLES15-SP3-15.3-0                 | SLES15-SP3-15.3-0              | No      | ----      | ----
3 | Server-Applications-Module_15.3-0 | sle-module-server-applications | No      | ----      | ----
```

As you can see, the internal alias is a better name than the name itself.

## Fix

- During installation the names are changed to the product name in the repository
- The reason is that normal repositories are added with name "Repository" and then after refreshing the repository the name is changed. However with Full medium the repositories already have a good name from the beginning, it is read from the `/media.1/products` index file.
- So the repository rename should be skipped if it is not the default name

repo_names_fixed_sp3

```
# zypper lr
Repository priorities are without effect. All enabled repositories share the same priority.

# | Alias                             | Name                              | Enabled | GPG Check | Refresh
--+-----------------------------------+-----------------------------------+---------+-----------+--------
1 | Basesystem-Module_15.3-0          | Basesystem-Module 15.3-0          | No      | ----      | ----
2 | SLES15-SP3-15.3-0                 | SLES15-SP3-15.3-0                 | No      | ----      | ----
3 | Server-Applications-Module_15.3-0 | Server-Applications-Module 15.3-0 | No      | ----      | ----
```
With the fix the repository names look much better.

## Tests

- Tested manually during installation
- Tested also in an installed system
- Added unit test

### DIFF
--- a/src/modules/Packages.rb
+++ b/src/modules/Packages.rb
@@ -1515,6 +1515,20 @@ module Yast
       true
     end
 
+    # Fallback repository name
+    # @return [String] translated name
+    def fallback_name
+      _("Repository")
+    end
+
+    # Is the repository name a fallback name?
+    # @param name [String] name of the repository
+    # @return [Boolean] true if it is a fallback name
+    def is_fallback_name?(name)
+      # to make the names unique a number might be appended at the end
+      name == fallback_name || name =~ /^#{Regexp.escape(fallback_name)}_\d+$/
+    end
+
     # Adjusts repository name according to a first product found on the media.
     #
     # @param [Fixnum] src_id repository ID
@@ -1527,6 +1541,13 @@ module Yast
       if src_id.nil? || Ops.less_than(src_id, 0)
         Builtins.y2error("Wrong source ID: %1", src_id)
         return nil
+      end
+
+      # change only the fallback name, otherwise the name already contains a product
+      repo_name = Pkg.SourceGeneralData(src_id)["raw_name"]
+      if !is_fallback_name?(repo_name)
+        log.info("Keeping repository name #{repo_name.inspect} unchanged")
+        return false
       end
 
       Builtins.y2milestone("Trying to get repository name from products")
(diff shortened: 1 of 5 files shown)


src/modules/Packages.rb
```diff
@@ -1515,6 +1515,20 @@ def Initialize_BaseInit(show_popup, base_url, log_url)
       true
     end
 
+    # Fallback repository name
+    # @return [String] translated name
+    def fallback_name
+      _("Repository")
```
jreidinger, line 1521: is it really good idea to have it localized? Other repos are also not localized, not?

src/modules/Packages.rb
```diff
@@ -1550,6 +1571,7 @@ def AdjustSourcePropertiesAccordingToProduct(src_id)
         sources_set = []
         Builtins.foreach(sources_got) do |one_source|
           if Ops.get_integer(one_source, "SrcId", -1) == src_id
+            log.info("Updating repository name from \"#{repo_name}\" to \"#{new_name}\"")
```
jreidinger, line 1574: another option is to use inspect and not need double quote. It also helps if name is nil.
```ruby
irb(main):001:0> nil.to_s
=> ""
irb(main):002:0> nil.inspect
=> "nil"
```

lslezak: Initially I used `#inspect` but then Rubocop complained about a too long line. :smiley:
So I changed that to a shorter form. And `nil` should not happen here, pkg-bindings would already complain earlier...

src/modules/Packages.rb
lslezak: I'm not sure, but normally the fallback should not be used...

jreidinger (approved):

PR #597, opened 2021-12-10
Ruby3

jreidinger:
PReparation for ruby3 and fix wrong fix for handling iso schema.

### DIFF
--- a/src/lib/y2packager/clients/inst_productsources.rb
+++ b/src/lib/y2packager/clients/inst_productsources.rb
@@ -2,7 +2,7 @@ require "yast"
 require "yast2/hw_detection"
 
 require "shellwords"
-require "uri"
+require "cgi"
 
 Yast.import "UI"
 Yast.import "Pkg"
@@ -354,7 +354,8 @@ module Yast
       return url_string if url_string.nil? || url_string == ""
 
       # unescape it
-      url_string = URI.unescape(url_string)
+      url_string = CGI.unescape(url_string)
+      log.info "unescaped url_string #{url_string.inspect}"
 
       # removing all slashes at the end of the url
       url_string.sub(/(\/)+$/, "")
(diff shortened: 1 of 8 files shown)


src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -354,7 +354,8 @@ def NormalizeURL(url_string)
       return url_string if url_string.nil? || url_string == ""
 
       # unescape it
-      url_string = URI.unescape(url_string)
+      url_string = CGI.unescape(url_string)
+      log.info "unescaped url_string #{url_string.inspect}"
```
lslezak, line 358: URLs might potentially contain sensitive data like password, URLs should be always filtered using the `Yast::URL.HidePassword(url)` call when logging them.

src/lib/y2packager/self_update_addon_repo.rb
```diff
@@ -88,8 +88,9 @@ def self.present?(path = REPO_PATH)
     # @return [Boolean] true on success, false if failed
     #
     def self.create_repo(path = REPO_PATH)
-      ret = Yast::Pkg.SourceCreateType("dir://#{URI.escape(path)}?alias=SelfUpdate0",
-        "", "Plaindir")
+      ret = Yast::Pkg.SourceCreateType(
+        "dir://#{Yast::URLRecode.EscapePath(path)}?alias=SelfUpdate0", "", "Plaindir"
```
lslezak, line 92: What's the difference to `path.split("/").map { |d| CGI.escape(d) }.join("/")` used below?

lslezak (commented):
Password logging

src/lib/y2packager/self_update_addon_repo.rb
jreidinger: well, this is dedicated method just for paths. Difference to below method. To be honest I am not sure which one is better as both mangle if it gets non path part of URL and encode it wrongly.

jreidinger: also that split is used in test code where I do not want too much rely on other code. But here in "production" code I would like to delete escaping.

lslezak (approved):

PR #598, opened 2022-01-12
New rubocop

jreidinger:


src/clients/sw_single.rb
```diff
@@ -623,26 +625,24 @@ def StartSWSingle
                 )
               end
               force_restart = true
-            elsif result == :online_search
+            when :online_search
               WFM.CallFunction("online_search", [:sw_single_mode])
               force_restart = true
-            elsif result == :webpin
+            when :webpin
```
lslezak, line 631: Webpin has been dropped long time ago, you can delete it completely.

src/lib/y2packager/clients/inst_productsources.rb
```diff
@@ -1316,14 +1312,13 @@ def SourcesDialog
...
             end
-          elsif @script_called_from_another
-            break
-          elsif Popup.ContinueCancelHeadline(
-            # TRANSLATORS: popup header
-            _("Aborting Configuration of Online Repository"),
-            # TRANSLATORS: popup question
-            _("Are you sure you want to abort the configuration?")
-          )
+          elsif @script_called_from_another ||
```
lslezak, line 1315: Is this correct? Previously the popup was not called when `@script_called_from_another` was true...

src/lib/y2packager/widgets/simple_language_selection.rb
```diff
@@ -95,7 +96,7 @@ def items
         @items = languages.map do |lang|
           [lang, LanguageTag.new(lang).name(lang_map_cache: lmap)]
         end
-        @items.reject! { |_lang, name| name.nil? }
+        @items.compact!
```
lslezak, line 99: I'm not sure if that works the same way as before:
```
irb(main):001:0> [ ["foo", nil] ].compact
=> [["foo", nil]]
```

src/modules/Packages.rb
```diff
@@ -1555,12 +1553,10 @@ def AdjustSourcePropertiesAccordingToProduct(src_id)
       new_name = nil
       Builtins.foreach(all_products) do |one_product|
         # source ID matches
-        if one_product.source == src_id
-          if one_product.name != ""
-            new_name = one_product.name
-            Builtins.y2milestone("Product name found: %1", new_name)
-            raise Break
-          end
+        if one_product.source == src_id && (one_product.name != "")
```
lslezak, line 1556: Are the parenthesis needed?

src/modules/SourceDialogs.rb
```diff
@@ -1428,23 +1427,14 @@ def IsoWidget
...
-      when :ftp
-        UI.SetFocus(:server)
-      when :http
-        UI.SetFocus(:server)
-      when :https
-        UI.SetFocus(:server)
-      when :samba
-        UI.SetFocus(:server)
-      end
+      UI.SetFocus(:server) if [:ftp, :http, :https, :samba].include?(server_type)
```
lslezak, line 1437: :+1:

src/modules/SourceDialogs.rb
```diff
@@ -2129,17 +2119,17 @@ def SelectWidgetHelp
...
       help_text = _(
         "<p><big><b>Media Type</b></big><br>\n" \
-          "The software repository can be located on CD, on a network server,\n" \
-          "or on the hard disk.</p>"
+        "The software repository can be located on CD, on a network server,\n" \
+        "or on the hard disk.</p>"
       )
 
       # help, continued
       help_text = Ops.add(
```
lslezak, line 2127: Could we get rid of those annoying `Ops.add`?

src/modules/SpaceCalculation.rb
```diff
@@ -825,27 +806,25 @@ def GetPartitionWarning
...
-          else
-            Builtins.add(
-              message,
-              "\n" +
-                # popup message
-                _("Deselect some packages.")
-            )
-          end
+      if Ops.greater_than(Builtins.size(message),
+        0) && (ProductFeatures.GetFeature("software", "selection_type") == :auto)
```
lslezak, line 810: Um, this leading `0)` looks strange, should be on the previous line...

test/packages_test.rb
```diff
@@ -366,7 +366,7 @@ def package(properties = {})
 
       expect(Yast::Y2Logger.instance).to receive(:info) do |msg|
         expect(msg).to match(
-          /(transaction\sstatus\s[begin|end]|(locked)?resolvables\s
+          /(transaction\sstatus\s[begin|d]|(locked)?resolvables\s
```
lslezak, line 369: This looks strange, is it correct?

lslezak (commented):
Some questions and comments

src/lib/y2packager/clients/inst_productsources.rb
jreidinger: it is not call even now. || and && is short evaluated:

```ruby
irb(main):002:0> true || puts("fail")
=> true
```

src/lib/y2packager/widgets/simple_language_selection.rb
jreidinger: yeah, here it is bug. As it expects it is hash. For hash it works, for array of arrays not.

src/modules/Packages.rb
jreidinger: probably not. Let me try what rubocop says, but it is in general now more strict about operator precedence and prefers parentheses to avoid confusion

src/modules/SourceDialogs.rb
jreidinger: yes, it is of course possible. But question is where we will stop :)

src/modules/SpaceCalculation.rb
jreidinger: agreed

test/packages_test.rb
jreidinger: sadly it looks previous regexp is wrong `[begin|d]` is same as `[bdegin|]` it is list of chars...and `e` `n` is already there.

src/clients/sw_single.rb
jreidinger: good. I always enjoy killing code.

lslezak (approved):

PR #599, opened 2022-01-14
Enable RSpec verifying doubles

lslezak:
- It found some problems in the tests which have been fixed
- Tested both locally (using the real YaST modules) and in `rake osc:build` using the new mocks
- All works fine

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -45,7 +45,36 @@ stub_module("FTP")
 stub_module("HTTP")
 stub_module("NtpClient")
 
-# helper for defining missing YaST modules
+if ENV["COVERAGE"]
+  require "simplecov"
+  SimpleCov.start do
+    add_filter "/test/"
+  end
+
+  # track all ruby files under src
+  SimpleCov.track_files("#{srcdir}/**/*.rb")
+
+  # additionally use the LCOV format for on-line code coverage reporting at CI
+  if ENV["CI"] || ENV["COVERAGE_LCOV"]
+    require "simplecov-lcov"
+
+    SimpleCov::Formatter::LcovFormatter.config do |c|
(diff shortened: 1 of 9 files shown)


test/test_helper.rb
```diff
@@ -68,6 +72,54 @@ def stub_module(name)
...
+# of the imported modules is correctly counted
+def define_if_missing(name, &block)
+  # try loading the module, it might be present in the system (running locally
+  # or in GitHub Actions), mock it only when missing (e.g. in OBS build)
+  Yast.import name
+  puts "Found module Yast::#{name}"
+rescue NameError
+  warn "Mocking the Yast::#{name} module completely"
+  block.call
+end
```
jreidinger, line 86: maybe we can add it to yast/rspec in ruby bindings? I think we face it in more modules and we handle it quite inconsistently.

jreidinger (approved):

test/test_helper.rb
```diff
@@ -68,6 +60,29 @@ def stub_module(name)
...
+Yast::RSpec::Helpers.define_yast_module("InstFunctions", methods: [:second_stage_required?])
+
+Yast::RSpec::Helpers.define_yast_module("Language", methods: [:language]) do
+  # @param _lang [String]
+  # @return [Boolean]
+  def supported_language?(_lang); end
+
+  # @param _force [Boolean]
+  # @return [Hash<String,Array>]
+  def GetLanguagesMap(_force); end
```
jreidinger, line 83: with change how method is defined, this won't be needed.

lslezak: no, see https://github.com/yast/yast-ruby-bindings/pull/281#discussion_r786956702

jreidinger (approved):

PR #603, opened 2022-02-23
Remake some icons

hellcp:
Also fixes a bug with checkmedia icon not rendering properly with qt svg renderer


dgdavid:
Oh, sorry. Looks like this has been out of our radar for quite long time already. Is the PR still valid?

dgdavid:
> Is the PR still valid?

If so, I guess that a version bump and changelog is needed. What do you think @lslezak?

hellcp:
Yes

jreidinger (approved):

PR #606, opened 2022-03-15
do not keep file handle to repo metadata open accidentally (bsc#1196061)

wfeldt:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1196061

A file handle to repo meta data (`*-primary.xml.gz`) is accidentally kept open preventing the detaching of installation media.

## Solution

Close file explicitly when no longer needed.


lslezak (approved):
LGTM
What about version bump and *.changes?

lslezak (approved):

PR #609, opened 2022-03-24
Show Download Progress

shundhammer:
## Target Branch / Product

**_This is for SLE-15-SP4._**

Version for master: _TBD_

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1195608

## Trello

https://trello.com/c/38oLzang/

## Problem

During package installation (only in the installed system, not during installation) there was no visual feedback while packages were being downloaded, only when all downloads were finished and it started actually installing the downloaded packages: The user saw only a progress bar stuck at 0% for minutes while the download was in progress.

During system installation, this was not a problem since download and installation is now (since SLE-15 SP4 / Leap 15.4) done in parallel.

## Cause

After the progress reporting was greatly simplified to enable parallel actions in libzypp, there was only one single progress bar, no lists of current actions (which had included downloads).

## Fix

Now displaying both the downloads _and_ the installed / upgraded / removed packages in that single progress bar.

- Total: Expected (unpacked) install sizes of all packages + expected (packed) download sizes

- Current: Finished install sizes + finished download sizes + partial download sizes

## Video

https://user-images.githubusercontent.com/11538225/161561914-ec78f295-236d-4bb3-9e57-2fe2f98974cd.mp4

## Rationale

Of course this is a little apples vs. oranges: Downloading 100 MB of RPMs rarely takes the same time as installing 100 MB of RPMs. But this makes the progress bar move during all waiting times, so the user can see that there is progress.

While doing similar operations on the command line, many seasoned users do

```
sudo zypper refresh
sudo zypper dup --download-only -y
```

And only when this is finished:

```
sudo zypper dup
```

This makes sure that this is a transaction, and it doesn't get stuck in the middle because of some packages that could not be downloaded. The current policy of the YaST software module in the installed system does very much the same: First, all required packages are downloaded, and when that is finished, the actual install / upgrade / remove operations are started.

Depending on the user's Internet connection, the download phase may take considerably longer than the installation phase, so the progress bar will very likely move quicker during the second phase. But it is very unlikely that any user will complain when it speeds up during the process; that will be very welcome by most users.

## Parallel Download and Installation

During system installation, we are now using a new libzypp mode that does downloading and installing packages in parallel, so the code needs to take that into account.

The first naive approach was to just check for `Mode.normal` (i.e. only in the installed system, not during system installation / system upgrade / AutoYaST installation). That works right now, but who knows when we will start using that new mode also in other scenarios.

So the code now keeps track in the callbacks (`DownloadStart()`, `DownloadProgress()`, `DownloadEnd()`, `PkgInstallStart()`, ...) if there is any indication that a package starts being installed while another one is downloaded, and changes to the _parallel_download_ mode, i.e. the progress bar only takes the package installation into account and not the downloads anymore; i.e. it auto-adapts.

The initial default is still based on `Mode.normal`, though.

## Other Changes

- Cleaned up that `PackageSlideShow` module:
  - Removed code duplication (initializing the member variables)
  - Removed redundant member variables: Some counter variables were really only the size of lists (packages installed, packages updated, ...). It never made any sense to duplicate that information. Now using `@xxx_list.size` instead wherever needed.
  - Clearly renamed the download callbacks to indicate what they do.
  - Factored out the very uncommon `DownloadError()` case into a separate method for clearer control flow; the uncommon case should not obscure the common one (download finished normally), much less with _always_ passing 2 (!) out of 3 method parameters for the uncommon error case.
  - Removed most YCP zombies (`Ops.Add()`, seriously?!)

## Installer Self-Update?

This affects only the installed system, not the system installation or upgrade, so an installer self-update is not needed.

## Related PR

- https://github.com/yast/yast-yast2/pull/1250

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr  6 14:00:54 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Show package downloads in the global progress bar during package
+  installation (bsc#1195608)
+- 4.4.28
+
+-------------------------------------------------------------------
 Thu Mar 31 15:52:11 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
 
 - Fixed regression in repository alias name for add-ons (bsc#1193214)
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -3,6 +3,7 @@ Wed Apr  6 14:00:54 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Show package downloads in the global progress bar during package
   installation (bsc#1195608)
+  PR: https://github.com/yast/yast-packager/pull/609
 - 4.4.28
 
 -------------------------------------------------------------------
--- a/src/modules/PackageSlideShow.rb
+++ b/src/modules/PackageSlideShow.rb
@@ -223,7 +223,7 @@ module Yast
     # @deprecated Misleading method name. For API backwards compatibility.
     #
     def DisplayGlobalProgress
-      log.warning "DEPRECATED. Use UpdateTotalProgressText() instead."
+      log.warn "DEPRECATED. Use UpdateTotalProgressText() instead."
       UpdateTotalProgressText()
     end
 
--- a/src/modules/PackageSlideShow.rb
+++ b/src/modules/PackageSlideShow.rb
@@ -130,8 +130,8 @@ module Yast
       @expected_total_download_size = Packages.CountSizeToBeDownloaded
       @init_pkg_data_complete = true
 
-      log.info "Total size to install: #{@total_size_to_install}"
-      log.info "Expected total download size: #{@expected_total_download_size}"
+      log.info "Total size to install: #{String.FormatSize(@total_size_to_install)}"
+      log.info "Expected download size: #{String.FormatSize(@expected_total_download_size)}"
       log.info "Parallel download (initial): #{parallel_download?}"
       nil
     end
@@ -189,12 +189,12 @@ module Yast
       else
         current = TotalInstalledSize()
         current += CurrentDownloadSize() unless parallel_download?
-        log.debug "Current: #{current} of #{total_size}"
-        total_progress = (100.0 * current / total_size).round
+        log.debug "Current: #{String.FormatSize(current)} of #{String.FormatSize(total_size)}"
+        total_progress = 100.0 * current / total_size
       end
 
-      log.info "Total progress: #{total_progress}%"
-      SlideShow.StageProgress(total_progress, nil)
+      log.debug "Total progress: #{total_progress.round(2)}%"
+      SlideShow.StageProgress(total_progress.round, nil)
     end
 
     # Calculate the size of the current downloads from finished downloads and
@@ -253,7 +253,7 @@ module Yast
     # Notification when download of a package starts
     def DownloadStart(pkg_name, download_size)
       @active_downloads += 1
-      log.info "DownloadStart #{pkg_name} size: #{download_size}"
+      log.info "Starting download of #{pkg_name} (#{String.FormatSize(download_size)})"
       log.info "active downloads: #{@active_downloads}" if @active_downloads > 1
       @current_download_pkg_name = pkg_name
       @current_download_pkg_size = download_size
@@ -269,7 +269,7 @@ module Yast
 
     # Update the download progress for the current package
     def DownloadProgress(pkg_percent)
-      log.info "#{@current_download_pkg_name}: #{pkg_percent}%"
+      log.debug "#{@current_download_pkg_name}: #{pkg_percent}%"
       @current_download_pkg_percent = pkg_percent
       return if parallel_download?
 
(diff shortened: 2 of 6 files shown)


src/modules/PackageSlideShow.rb
```diff
@@ -125,60 +121,90 @@ def InitPkgData(force)
...
+      nil
     end
 
-    # Switch unit to seconds if necessary and recalc everything accordingly.
-    # @return true if just switched from sizes to seconds, false otherwise
-    #
-    def SwitchToSecondsIfNecessary
-      false
-    end
+    # Check if package are downloaded in parallel to being installed.
```
joseivanlopez, line 139: Typos: if packages are downloaded in parallel to be installed

src/modules/PackageSlideShow.rb
```diff
@@ -188,150 +214,203 @@ def UpdateTotalProgress(_silent_check)
...
+    def UpdateDownloadProgressText
+      SlideShow.SetGlobalProgressLabel(
+        _("Downloading...") +
+        Builtins.sformat(
+          # TRANSLATORS: This is about a remaining download size.
+          # %1 is the remaining size with a unit (kiB, MiB, GiB etc.),
+          # %2 the total download size, also with a unit.
+          _(" (Remaining: %1 of %2)"),
+          String.FormatSize(@expected_total_download_size - CurrentDownloadSize()),
+          String.FormatSize(@expected_total_download_size)
```
joseivanlopez, line 246: As commented in IRC, I am used to counter up. For example, browsers usually have: "Downloading 8/489 MiB". Somehow, the current version with remaining makes me mentally process it. What do you think?

src/modules/PackageSlideShow.rb
```diff
@@ -188,150 +214,203 @@ def UpdateTotalProgress(_silent_check)
...
-          Builtins.sformat(
-            _(" (Remaining: %1%2 packages)"),
-            rem_string,
-            @total_pkgs_to_install - @installed_packages - @updated_packages
-          )
+        SlideShow.CurrentStageDescription +
+        Builtins.sformat(
+          _(" (Remaining: %1%2 packages)"),
+          remaining_string,
+          @total_pkgs_to_install - installed_pkg - updated_pkg
```
joseivanlopez, line 267: And I have the same feeling here. IMHO, it would be more understandable to have "Installating: 8/342 packages". On the other hand, a counter up is a mathematical representation of the pogress bar. So, maybe the remainder value is giving more information. I don't know.

src/modules/PackageSlideShow.rb
```diff
@@ -188,150 +214,203 @@ def UpdateTotalProgress(_silent_check)
...
-    # @param [Integer] _pkg_size    package size in bytes
-    # @param [Boolean] deleting    Flag: deleting (true) or installing (false) package?
+    # @param [Integer] _pkg_percent percent of progress of this package
     #
-    def SlideDisplayStart(pkg_name, _pkg_location, _pkg_summary, _pkg_size, deleting)
-      @updating = Pkg.PkgInstalled(pkg_name) if !deleting
-      # Update global progress bar
-      DisplayGlobalProgress()
-
+    def PkgInstallProgress(_pkg_percent)
```
joseivanlopez, line 363: Coud this method be removed?

shundhammer: Could, yes, at the cost of clarity since the libzypp callbacks one call level above do have such a method. This is still here for future use and for orientation of somebody reading the code later who will for sure wonder why this mirrors so many of the other callbacks (from the `SlideShowCallbacks` level), but not this one.

But you are right, this deserves a comment in the code.

src/modules/PackageSlideShow.rb
shundhammer: No, it's really meant like this: It's the download and the installation (of other packages) that are in parallel.

src/modules/PackageSlideShow.rb
```diff
@@ -188,150 +214,203 @@ def UpdateTotalProgress(_silent_check)
       UpdateTotalProgressValue()
     end
 
-    # Progress display update
-    # This is called via the packager's progress callbacks.
+    # For API backwards compatibility. DEPRECATED.
```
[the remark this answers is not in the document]
shundhammer: OK. Will do.

src/modules/PackageSlideShow.rb
shundhammer: Let's discuss this live later today.

I don't have a strong opionion one way or the other; It was just like that before, and now we have two locations with very similar messages. We should be consistent.

src/modules/PackageSlideShow.rb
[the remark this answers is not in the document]
shundhammer: Done.

src/modules/PackageSlideShow.rb
shundhammer: Added an explanatory comment.

package/yast2-packager.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Wed Apr  6 14:00:54 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Show package downloads in the global progress bar during package
+  installation (bsc#1195608)
+  PR: https://github.com/yast/yast-packager/pull/609
```
joseivanlopez, line 6: Was the PR URL intentionally added?

joseivanlopez (approved):
LGTM

package/yast2-packager.changes
shundhammer: Yes, so it's easier to get to the PR with more explanations. I started doing that some weeks ago after I had to hunt down a PR.

PR #612, opened 2022-04-07
Show Download Progress [master]

shundhammer:
## Target Branch / Product

**_This is for master / Factory / Tumbleweed._**

This merges https://github.com/yast/yast-packager/pull/609 to master / Factory / Tumbleweed.

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1195608

## Trello

https://trello.com/c/38oLzang/

## Problem

During package installation (only in the installed system, not during installation) there was no visual feedback while packages were being downloaded, only when all downloads were finished and it started actually installing the downloaded packages: The user saw only a progress bar stuck at 0% for minutes while the download was in progress.

During system installation, this was not a problem since download and installation is now (since SLE-15 SP4 / Leap 15.4) done in parallel.

## Cause

After the progress reporting was greatly simplified to enable parallel actions in libzypp, there was only one single progress bar, no lists of current actions (which had included downloads).

## Fix

Now displaying both the downloads _and_ the installed / upgraded / removed packages in that single progress bar.

- Total: Expected (unpacked) install sizes of all packages + expected (packed) download sizes

- Current: Finished install sizes + finished download sizes + partial download sizes

## Video

https://user-images.githubusercontent.com/11538225/161561914-ec78f295-236d-4bb3-9e57-2fe2f98974cd.mp4

## Rationale

Of course this is a little apples vs. oranges: Downloading 100 MB of RPMs rarely takes the same time as installing 100 MB of RPMs. But this makes the progress bar move during all waiting times, so the user can see that there is progress.

While doing similar operations on the command line, many seasoned users do

```
sudo zypper refresh
sudo zypper dup --download-only -y
```

And only when this is finished:

```
sudo zypper dup
```

This makes sure that this is a transaction, and it doesn't get stuck in the middle because of some packages that could not be downloaded. The current policy of the YaST software module in the installed system does very much the same: First, all required packages are downloaded, and when that is finished, the actual install / upgrade / remove operations are started.

Depending on the user's Internet connection, the download phase may take considerably longer than the installation phase, so the progress bar will very likely move quicker during the second phase. But it is very unlikely that any user will complain when it speeds up during the process; that will be very welcome by most users.

## Parallel Download and Installation

During system installation, we are now using a new libzypp mode that does downloading and installing packages in parallel, so the code needs to take that into account.

The first naive approach was to just check for `Mode.normal` (i.e. only in the installed system, not during system installation / system upgrade / AutoYaST installation). That works right now, but who knows when we will start using that new mode also in other scenarios.

So the code now keeps track in the callbacks (`DownloadStart()`, `DownloadProgress()`, `DownloadEnd()`, `PkgInstallStart()`, ...) if there is any indication that a package starts being installed while another one is downloaded, and changes to the _parallel_download_ mode, i.e. the progress bar only takes the package installation into account and not the downloads anymore; i.e. it auto-adapts.

The initial default is still based on `Mode.normal`, though.

## Other Changes

- Cleaned up that `PackageSlideShow` module:
  - Removed code duplication (initializing the member variables)
  - Removed redundant member variables: Some counter variables were really only the size of lists (packages installed, packages updated, ...). It never made any sense to duplicate that information. Now using `@xxx_list.size` instead wherever needed.
  - Clearly renamed the download callbacks to indicate what they do.
  - Factored out the very uncommon `DownloadError()` case into a separate method for clearer control flow; the uncommon case should not obscure the common one (download finished normally), much less with _always_ passing 2 (!) out of 3 method parameters for the uncommon error case.
  - Removed most YCP zombies (`Ops.Add()`, seriously?!)

## Related PR

- https://github.com/yast/yast-yast2/pull/1250

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Apr  7 11:28:48 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Show package downloads in the global progress bar during package
+  installation (bsc#1195608)
+  PR: https://github.com/yast/yast-packager/pull/612
+- 4.5.1
+
+-------------------------------------------------------------------
 Wed Apr 06 13:24:58 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
 
 - Bump version to 4.5.0 (#bsc1198109)
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.5.0
+Version:        4.5.1
 Release:        0
 Summary:        YaST2 - Package Library
 License:        GPL-2.0-or-later
--- a/src/modules/PackageSlideShow.rb
+++ b/src/modules/PackageSlideShow.rb
@@ -16,66 +16,62 @@ module Yast
       Yast.import "SlideShow"
       Yast.import "String"
       Yast.import "Mode"
+      Yast.import "Packages"
 
-      @inst_src_names = [] # a list of strings identifying each repository
-      @total_size_installed = 0
-      @total_size_to_install = 0
+      init_member_vars
+    end
 
+    def init_member_vars
       @init_pkg_data_complete = false
 
-      # package summary
-      # package counters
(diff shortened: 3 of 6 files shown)


joseivanlopez (approved):
LGTM

PR #613, opened 2022-04-26
Don't Rely on /etc/install.inf being Available [15-SP4]

shundhammer:
## Target Branch / Product

**This is for SLE-15-SP4.** A port to master will follow.

## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1198560

## Trello

https://trello.com/c/AXWDv2jh/

## Problem

When migrating from Leap 15.4 to SLE-15 SP4 on the installed system, YaST crashes with an exception _"The installation URL is not set"_ in https://github.com/yast/yast-packager/blob/master/src/lib/y2packager/installation_medium.rb#L48-L50 .

## Cause

It tried to read `/etc/install.inf` (the config file that _linuxrc_ writes for YaST to consume) to check if the installation media has several products or just one. But in this scenario, the migration started from the installed system, not from an inst-sys on installation media; so there was no `/etc/install.inf`.

## Fix

Check if there is an `install.inf`; more precisely, if any data from that could be read, i.e. if there was _any_ content at all from such a file.

If there is none, read the products from libzypp: In this scenario, the old repos from the system that we want to migrate away from were just removed, and the ones from the new one (that we want to migrate to) were added. Those repos are what we need here.

Notice that in this scenario it does _not_ make sense to read a repo URL from a `control.xml` file: While there is one in `/etc/YaST2/control.xml`, that is not the default path on installation media, so it won't be found; and, worse, that file contains the URL of the _old_ repo, the system that we are migrating away from.

## Related PR

PR for master: _TBD_

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Apr 27 13:31:36 UTC 2022 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Don't rely on install.inf availability #(bsc#1198560)
+- 4.4.30
+
+-------------------------------------------------------------------
 Wed Apr 27 11:24:44 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
 
 - Fixed migration summary in Leap -> SLES migration (bsc#1198562)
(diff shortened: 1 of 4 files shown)


lslezak (approved):
LGTM

PR #619, opened 2022-05-17
Fix package counters during installation (boo#1199621)

imobachgs:
## Problem

The count of remaining packages to install is not updated.

- [bsc#1199621](https://bugzilla.opensuse.org/show_bug.cgi?id=1199621)

## Solution

Keep the installed/updated/removed packages counters in separate variables. Alternatively, we could just update always the lists, but I guess the original idea was to not keep long lists of packages that are not going to be used. Unlike during installation/upgrade, in `Mode.normal`, the list is displayed at the end of the process.

## Testing

- *Tested manually*

## Screencast

installed-packages-counter


shundhammer (approved):
I just tested this with the latest TW and Leap 15.4 images, and I can confirm that this works.

PR #623, opened 2022-09-16
Do not call [Install done] on aborted packages

yecril71pl:
Fix [Bug 1203302](https://bugzilla.opensuse.org/show_bug.cgi?id=1203302)

## Problem

The slide show used by the package downloader does not recognise the return code for aborted packages and treats them as installed.

## Solution

Instead of checking that the return code is not RETRY, require it to be explicitly INSTALL.

## Testing

Currently beyond my possibilities, unfortunately.

## Screenshots

negative remaining


shundhammer (approved):
LGTM

Thank you for your contribution!
I'll create a follow-up PR to bump the version number and for a change log entry.

PR #625, opened 2022-10-31
Escape spaces in the installation URL (bsc#1201816)

lslezak:
## Problem

- The URL from `install.inf` might contain unescaped spaces. Spaces are not allowed in URL and should be escaped.
- See https://bugzilla.suse.com/show_bug.cgi?id=1201816#c10, https://bugzilla.suse.com/show_bug.cgi?id=1201816#c26

## Solution

- As a quick and simple fix just replace all spaces with the percent encoded values
- This does not need any change in `linuxrc` and can be deployed as a self-update

## Testing

- Tested manually

### DIFF
--- a/src/modules/InstURL.rb
+++ b/src/modules/InstURL.rb
@@ -157,6 +157,9 @@ module Yast
       @installInf2Url = add_extra_dir_to_url(@installInf2Url, extra_dir) unless extra_dir.empty?
       @installInf2Url = add_ssl_verify_no_to_url(@installInf2Url) unless SSLVerificationEnabled()
 
+      # escape spaces otherwise it would be invalid and rejected by libzypp
+      @installInf2Url.gsub!(" ", "%20")
+
       log.info "Using install URL: #{URL.HidePassword(@installInf2Url)}"
       @installInf2Url
     end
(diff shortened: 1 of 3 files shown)


lkocman (approved):
looks valid

src/modules/InstURL.rb
```diff
@@ -157,6 +157,9 @@ def installInf2Url(extra_dir = "")
       @installInf2Url = add_extra_dir_to_url(@installInf2Url, extra_dir) unless extra_dir.empty?
       @installInf2Url = add_ssl_verify_no_to_url(@installInf2Url) unless SSLVerificationEnabled()
 
+      # escape spaces otherwise it would be invalid and rejected by libzypp
+      @installInf2Url.gsub!(" ", "%20")
```
jreidinger, line 161: is it already stripped? or can it contain space at start and end? if so, then I think we should strip it at first and then do escaping.

jreidinger (approved):
just question to be sure, but if it cannot contain space at start and end, then LGTM

src/modules/InstURL.rb
lslezak: It should not contain any beginning or trailing space, but you are right, it would be better to remove it.

PR #627, opened 2022-11-02
Fixed adding an empty repository at upgrade (bsc#1204399)

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1204399
- After upgrading a registered SLE system a strange empty repository is present in the system (the name is like `cd-3494b1a8`, created as a random unique name)

## Details

- It turned out that it only happens when using the *Full* (offline) medium for upgrading a *registered* system
- The empty repository is actually the repository from the root of the Full medium (the real repositories are in subdirectoriees) 
- The code actually already contains a condition which should prevent from this problem ([this code](https://github.com/yast/yast-packager/blob/d1be27c5ea5bd061bc2a368c121127122231e383/src/modules/Packages.rb#L1666-L1669))
- But when the Full medium is used in a registered system it should behave like the Online medium. This is achieved by explicitly switching the detected medium type from Full to Online ([this code](https://github.com/yast/yast-registration/blob/431d937b78c209c0d35ce911a98f9f05978e77ca/src/lib/registration/ui/migration_repos_workflow.rb#L602-L605))

## Solution

- Remember the originally detected medium type so we can use it even when the type is changed later

## Testing

- Tested manually
- Without the patch a `cd-*` repository is added to the system
- With the patch no such repository is added, only the repositories from SCC are present

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Nov  2 12:35:52 UTC 2022 - Ladislav Slezák <lslezak@suse.cz>
+
+- Do not add an empty repository to the system when upgrading
+  a registered system using the Full installation medium
+  (bsc#1204399)
+- 4.4.33
+
+-------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


joseivanlopez (approved):
LGTM

PR #632, opened 2023-01-11
Media check fixes in containerized environment

lslezak:
- https://bugzilla.suse.com/show_bug.cgi?id=1206834
- Start browsing the ISO files from the `/host` directory  to show the host files, starting from the container root is confusing for users (see the bugreport)
- Fixed file readablity check to correctly use the specified ISO file (`.target_bash` runs in `/host` chroot, the file selected by UI file selector starts from `/`, moreover we can easily replace the external shell command by native Ruby code)

### DIFF
--- a/src/include/checkmedia/ui.rb
+++ b/src/include/checkmedia/ui.rb
@@ -23,6 +23,8 @@ require "shellwords"
 module Yast
   # All user interface functions.
   module CheckmediaUiInclude
+    include Yast::Logger
+
     def initialize_checkmedia_ui(_include_target)
       Yast.import "Pkg"
       Yast.import "UI"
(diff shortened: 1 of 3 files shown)


src/include/checkmedia/ui.rb
```diff
@@ -555,6 +552,18 @@ def MainDialog
       ret
     end
 
+    # try to read one byte from the medium to check whether it is readable
+    # @param file [String] file path
+    # @return [Boolean] returns `true` if the file is readable, `false` otherwise
+    def file_readable?(file)
+      File.open(file) do |f|
+        f.read(1)
+      end
```
jreidinger, line 561: you can use `File.read(file, 1)`

src/include/checkmedia/ui.rb
```diff
@@ -555,6 +552,18 @@ def MainDialog
...
+      true
+    rescue StandardError
+      false
```
jreidinger, line 564: I would at least log content of exception if we get some bug report. So we see how it failed.

PR #633, opened 2023-01-18
Fix URI to work with Ruby 3.2

mvidner:
## Problem

Context: Ruby 3.2 is coming to Tumbleweed but YaST fails to build. Staging:H.

This part of the failures is: `URI("dir:///foo").to_s` no longer returns `"dir:/foo"`

- https://bugzilla.suse.com/show_bug.cgi?id=1207239 
- https://trello.com/c/jRe5bWVE/3249-ostumbleweed-1206419-staging-ruby-keyword-parameters-yast-stack-fails-to-build-against-ruby-32

## Solution

- some of the failing tests are fixed in the `ZyppUrl` class in https://github.com/yast/yast-yast2/pull/1285
- the other half here: change the test expectation to allow for 3 slashes (here %-escaped) since zypp treats them the same anyway. And allow for 1 slash to continue building with older Rubies

## Testing

- it's all in existing unit tests

## Screenshots

N/A

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Feb 14 13:40:12 UTC 2023 - Martin Vidner <mvidner@suse.com>
+
+- Ruby 3.2: Change a test to treat dir:///foo equal to dir:/foo
+  (bsc#1207239)
+- 4.5.15
+
+-------------------------------------------------------------------
 Wed Feb  8 17:29:25 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
 
 - Prevent crash if nil dependencies instead of [] (bsc#1208068)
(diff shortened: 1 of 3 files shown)


mvidner:
This 1-slash form is compared to a 3-slash form which doesn't match.

```console
$ ruby -r yast -e 's = "dir:///foo"; Yast.import "Pkg"; p Yast::Pkg.ExpandedUrl(s)'
"dir:/foo"
```

We're entering C++ territory in Pkg and libzypp: https://github.com/yast/yast-pkg-bindings/blob/987c4933a9fa8935664bd35d08d4c5ce5d050e30/src/PkgFunctions.cc#L554

test/pkg_finish_test.rb
```diff
@@ -271,7 +271,7 @@
       context "and the products are also defined in the remote repos" do
         let(:remote_products) { [sles_product, sled_product] }
 
-        it "removes the fallback repository" do
+        it "removes the fallback repository" do # HERE 1/2
```
mvidner, line 274: Oops, debugging has snuck in

jreidinger (approved):

PR #634, opened 2023-01-26
Unescape the product name read from package Provides (bsc#1207276)

lslezak:
## Problem

- The product names in Provides is now percent escaped (as some other Provides values already do)
- https://bugzilla.suse.com/show_bug.cgi?id=1207276

## Notes

This is backward compatible, if escaping is not used it is still safe:

```
irb(main):002:0> URI.unescape("sle%2Dmodule%2Dweb%2Dscripting")
=> "sle-module-web-scripting"
irb(main):003:0> URI.unescape("sle-module-web-scripting")
=> "sle-module-web-scripting"
```


shundhammer (approved):
LGTM

PR #636, opened 2023-02-08
Prevent Crash if nil Product Dependencies instead of Empty List

shundhammer:
## Bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1208068

## Trello

https://trello.com/c/HVBMy1Qf/

## Problem

"Internal error" crash when the SUSE Manager add-on product was selected.

## Cause

The dependencies of that add-on product were most likely specified wrong on the SCC or on the repo side: `nil` instead of an empty list `[ ]`.

## Fix

Made the code more robust against this sort of problem: It will most likely happen again in the future. Added the Ruby `&.` selector which survives a `nil`, unlike the normal `.`. Of course then the resulting list also needs to be filtered for `nil` with `compact()` or `compact!()`.

## Related Problem

That add-on product should actually have dependencies, at least the SLE-15 base product.

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Feb  8 17:29:25 UTC 2023 - Stefan Hundhammer <shundhammer@suse.com>
+
+- Prevent crash if nil dependencies instead of [] (bsc#1208068)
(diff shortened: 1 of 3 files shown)


lslezak (approved):
LGTM, thanks!

PR #637, opened 2023-02-20
Fixed a crash when selecting depending products (bsc#1208421)

lslezak:
## Problem

- Fixes a crash when selecting the HA product
- https://bugzilla.suse.com/show_bug.cgi?id=1208421

## Solution

- The `uniq!` call returns `nil` if there is no duplicate in the Array
- The `compact!` call must not be chained, it needs to be called on the variable again

## Testing

- Added a new unit test


lslezak:
The Tumbleweed CI builds fail with `cannot load such file -- yast/rake`, I'll investigate the problem later, let's get merged this P1 fix ASAP.

jreidinger (approved):

PR #638, opened 2023-04-06
Fix tests

jreidinger:
## Problem

build failing at https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:G/yast2-packager/standard/x86_64

## Solution

Fix test to use proper string instead of Pathname.

## Testing

- *Tested manually*

### DIFF
--- a/src/lib/packager/product_patterns.rb
+++ b/src/lib/packager/product_patterns.rb
@@ -99,7 +99,6 @@ module Yast
         end
       end
 
-
       # commented out block debug call until issue is fixed: https://github.com/yast/yast-ruby-bindings/issues/290
       # log.debug { "Product #{product} depependencies: #{product_dependencies}" }
 
--- a/src/lib/packager/product_patterns.rb
+++ b/src/lib/packager/product_patterns.rb
@@ -99,7 +99,9 @@ module Yast
         end
       end
 
-      log.debug "Product #{product} depependencies: #{product_dependencies}"
+
+      # commented out block debug call until issue is fixed: https://github.com/yast/yast-ruby-bindings/issues/290
+      # log.debug { "Product #{product} depependencies: #{product_dependencies}" }
 
       product_dependencies
     end
@@ -131,7 +133,8 @@ module Yast
         provides << prov if prov
       end
 
-      log.debug "Collected provides dependencies: #{provides.inspect}"
+      # commented out block debug call until issue is fixed: https://github.com/yast/yast-ruby-bindings/issues/290
+      # log.debug { "Collected provides dependencies: #{provides.inspect}" }
 
       provides
     end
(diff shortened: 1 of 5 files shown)


src/lib/packager/product_patterns.rb
```diff
@@ -99,8 +99,6 @@ def dependencies(product)
         end
       end
 
-      log.debug "Product #{product} depependencies: #{product_dependencies}"
```
mvidner, line 102: You are smuggling in an undocumented change, and I have a better one to replace it :smile:

mvidner:
https://github.com/yast/yast-ruby-bindings/blob/9a50d95be89f3dfe81cd109e6bf89a5fbedc48eb/src/ruby/yast/y2logger.rb#L48 shows that we're using a standard logger, and the [Ruby documentation shows](https://docs.ruby-lang.org/en/master/Logger.html#class-Logger-label-Entries) that passing a block will avoid evaluating it when debugging is disabled.

So please use `log.debug { "nasty string #{mem_bomb}" }` instead

BTW that Ruby docs is so skillful in hiding the useful information, it sucks

jreidinger:
@mvidner sadly block form does not work, as we do not limit access to debug in logger level but on ruby-bindings level

try yourself in irb:
```
1 require "yast"
  2 
  3 include Yast::Logger
  4 
  5 log.debug { puts "eval"; "log content" }
```

mvidner:
Ah, good point. Then what about

1. make an Issue in ruby-bindings "log.debug { expensive } always evaluates the block"
2. leave the calls here, commented out, referencing that issue

jreidinger:
issue created https://github.com/yast/yast-ruby-bindings/issues/290

mvidner (commented):
Fixes so far look good, but what about
- the other failure "unexpected message :define_method with ("switch_to_richtext=")", do you need help?
- version + changelog?

test/addon_selector_test.rb
```diff
@@ -66,7 +66,8 @@
 
       it "does not display any popup" do
         # stub empty Yast::Popup so any method call would raise an exception
-        stub_const("Yast::Popup", double)
```
mvidner, line 69: Does that mean that any such `stub_const("Yast::Foo", double)` will fail or has bug? We have more

https://github.com/search?q=org%3Ayast+stub_const&type=code

mvidner: yast/yast-update/test/inst_update_partition_auto_test.rb
15:      stub_const("Yast::FileSystems", double)
yast/yast-installation/test/snapshots_finish_test.rb
10:    stub_const("Yast::StorageSnapper", double)

mvidner (approved):
Thanks!

I think there is still a chance that another fix will be needed in ruby-bindings because of the following, but I will not let that block this PR.

> Does that mean that any such `stub_const("Yast::Foo", double)` will fail or has bug? We have more: https://github.com/search?q=org%3Ayast+stub_const&type=code

mvidner:
For the record, the `stub_const` fix adjusts for the change done in https://github.com/yast/yast-ruby-bindings/pull/289

PR #642, opened 2023-09-20
[SLE-15-SP6] Properly save the repo file (bsc#1214135)

lslezak:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1214135
- When changing URL of a repository belonging to a service the service name is lost in the `*.repo` file
- The result is that the repository is not linked to the service and refreshing the service fails because there is a conflict between the "standalone" repository and the repository coming from the service

## Details

```console
# zypper install openSUSE-repos-Tumbleweed
# zypper refresh -s
** list the original repo file **
# cat /etc/zypp/repos.d/openSUSE\:repo-non-oss.repo 
[openSUSE:repo-non-oss]
name=repo-non-oss
enabled=1
autorefresh=1
baseurl=http://cdn.opensuse.org/tumbleweed//repo/non-oss
service=openSUSE
# yast2 repositories 
** change the URL of the repo-non-oss repository, just change the http:// protocol to https:// **
# cat /etc/zypp/repos.d/openSUSE\:repo-non-oss.repo 
[openSUSE:repo-non-oss]
name=repo-non-oss
enabled=1
autorefresh=1
baseurl=https://cdn.opensuse.org/tumbleweed//repo/non-oss
path=/
type=rpm-md
keeppackages=0
** the "service" line is missing here! **
# zypper ref -s
Refreshing service 'openSUSE'.
Adding repository 'repo-non-oss' ........................................[error]
Unexpected exception.
[openSUSE:repo-non-oss|http://cdn.opensuse.org/tumbleweed//repo/non-oss] Repository already exists.
** so then refreshing the service fails **
```
The problem is that the `service=openSUSE` line is lost. On the other hand it added some more lines, but these lines just contain the defaults, the meaning is the same as if they were missing.

## Solution

- Changing the URL is a bit tricky, in that case YaST actually deletes the old repository (to force removing all cached files) and creates a new repository with the new URL. Then YaST copies all attributes (enabled/disabled, priority,...) from the old repository to the new one.
- And here was the bug, YaST did not copy the "service" attribute so the new repository had the same name as the old one, but was not bound to the original service. For libzypp it looks like a standalone repository and when libzypp refreshed the service it found out that a conflicting repository already exists.
- Also see https://github.com/yast/yast-pkg-bindings/pull/181, this requires a small update in the pkg-bindings package

## Testing

- Tested manually

```console
# yast2 repositories
** change the URL of the repo-non-oss repository **
# cat /etc/zypp/repos.d/openSUSE\:repo-non-oss.repo 
[openSUSE:repo-non-oss]
name=repo-non-oss
enabled=1
autorefresh=1
baseurl=https://cdn.opensuse.org/tumbleweed//repo/non-oss
path=/
type=rpm-md
keeppackages=0
service=openSUSE
** the "service" line is correctly saved **
# zypper refresh -s
Refreshing service 'openSUSE'.
All services have been refreshed.
...
All repositories have been refreshed.
** refresh succeeded **
```

When changing a repository which does not belong to any service the `service` line is *not* added so there is no regression.

## Screenshots

- Additionally a warning is displayed when changing the repository URL
- The same warning is already displayed when changing other repository attributes (enabled/disabled, priority,...), it was missing when changing the URL
service_warning

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Wed Sep 20 09:08:59 UTC 2023 - Ladislav Slezák <lslezak@suse.com>
+
+- Properly save the repo file when changing URL of a repository
+  from a service (bsc#1214135)
+- Also display a warning when changing URL of a service repository,
+  the change might be lost after the service is refreshed
+- 4.6.4
+
+-------------------------------------------------------------------
 Fri Sep 01 19:57:03 UTC 2023 - Josef Reidinger <jreidinger@suse.com>
 
 - Branch package for SP6 (bsc#1208913)
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.6.3
+Version:        4.6.4
 Release:        0
 Summary:        YaST2 - Package Library
 License:        GPL-2.0-or-later
@@ -38,8 +38,8 @@ BuildRequires:  yast2-storage-ng >= 4.0.141
 #!BuildIgnore: yast2-packager
 # Replace PackageSystem with Package
 BuildRequires:  yast2 >= 4.4.38
-# raw_name
-BuildRequires:  yast2-pkg-bindings >= 4.2.8
+# "service" attribute in Pkg.SourceEditSet()
+BuildRequires:  yast2-pkg-bindings >= 4.6.4
 # yast/rspec/helpers.rb
 BuildRequires:  yast2-ruby-bindings >= 4.4.7
 # Augeas lenses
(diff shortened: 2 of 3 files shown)


jreidinger (approved):
approved, but that repositories clients is really scary.

lslezak:
> approved, but that repositories clients is really scary.

Yes, that's a legacy from the YCP times...

mvidner (approved):
Thanks for the detailed description

mlandres:
Frankly, this is IMO just a hotfix.

- Changes to a service repo are volatile (except enable/disable) and undone on the next service refresh.
- In general [# set the original properties](https://github.com/yast/yast-packager/pull/642/commits/a69c645b8767093aa90f22bec768e66eddc84ef4#diff-6b784bf0e6812e2a348e4243710f8359a6bc378e181869ad343d4bc10a42d5e1L1908) appears to still lose settings (gpg related ones, name) and it will fore sure lose future attributes we will apply (like e.g. metadata_expire and skip_if_unavailable).

The bug shows that you lose other attributes as well (like the storage location of the repo data), which fortunately does not matter as long as it's the standard `_alias_.repo` file.

Regarding zypp it should be sufficient to simply refresh the repo, even after changing the URL. If the repo was already loaded into the ResPool, unload-refresh-reload. You don't need to deal with the zypp caches. Repo URL changes (incl. changed valuas of repo vars used in a URL) are detected and rill cause the repo to be tagged as _refresh_needed_.

mlandres:
Maybe you can take copy of the original RepoInfo and writeback the data you know about, rather than creating a new one from scratch.

PR #658, opened 2025-01-06
Ruby 3.4 adaptation

mvidner:
## Problem

Hash#inspect output has changed
from `{:name=>"openSUSE", :version=>"13.1-1.10"}`
to   `{name: "openSUSE", version: "13.1-1.10"}`

the output goes to the log, is not parsed further

- https://bugzilla.suse.com/show_bug.cgi?id=1235099

## Solution

Relax the regex that checks what gets logged

## Testing

- Caught by existing unit tests

But you need to run them with both old (`rake osc:build`) and new Ruby: 

```sh
osc co openSUSE:Factory:Staging:L/yast2-packager
cd $_
rm *.tar.bz2
cp .../yast-packager/package/* .
osc ar .
osc build
```
## Screenshots

No

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Jan  6 10:45:35 UTC 2025 - Martin Vidner <mvidner@suse.com>
+
+- Adjust test for ruby-3.4 (bsc#1235099)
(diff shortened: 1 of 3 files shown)


jreidinger (approved):
lets merge it...but I am not so fan of having log lines regexp matching which makes test fragile.

PR #660, opened 2025-07-15
Mark strings from libsolv as UTF-8 (bsc#1245555)

mvidner:
## Problem

When selecting SLE-HA as add-on product, either during installation or via "yast2 add-on", an Internal Error pops up as soon as one clicks that product.

It is a combination of two factors:
1. the product description (coming from the Description field of `sle-ha-release.rpm`) is in English but it contains a non-ASCII character (a curly apostrophe)
2. YaST retrieves that string via `ruby-solv.rpm`, a Swig binding of `libsolv`, which returns ALL strings marked with `Encoding::ASCII_8BIT`, which means `BINARY`

When trying to combine two strings, one `UTF_8` and the other `ASCII_8BIT`, `Encoding::CompatibilityError` is raised, *unless* one of the strings happens to contain pure ASCII, which is why htis problem went undetected for so long.

- https://bugzilla.suse.com/show_bug.cgi?id=1245555 (L3)
- https://trello.com/c/cRMoDGxs

## Solution

For the solvable fields likely to contain non-ASCII text, Summary and Description, call `String#force_encoding` on the value retrieved from libsolv

## Testing <a name="description-testing"></a>

Tested manually

1. running as root "yast2 add-on" on my Leap 15.6
2. Add... Url: http://download.suse.de/install/SLP/SLE-15-SP7-Full-PublicRC-202504/x86_64/DVD1/ ... Next
3. Click the "SUSE Linux ... High Availability ..." line
4. the Internal Error happens (or doesn't, with the fix)

## Screenshots

No

### DIFF
--- a/package/yast2-packager.changes
+++ b/package/yast2-packager.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 15 14:57:19 UTC 2025 - Martin Vidner <mvidner@suse.com>
+
+- Fix Internal Error: Encoding::CompatibilityError when
+  adding SLE-HA as add-on product (bsc#1245555)
+- 4.6.10
+
+-------------------------------------------------------------------
 Tue Mar 12 13:43:26 UTC 2024 - Ladislav Slezák <lslezak@suse.com>
 
 - Change the product mapping API to be the same as in SP5 and older
--- a/package/yast2-packager.spec
+++ b/package/yast2-packager.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-packager
-Version:        4.6.9
+Version:        4.6.10
 Release:        0
 Summary:        YaST2 - Package Library
 License:        GPL-2.0-or-later


jreidinger (approved):

lslezak (approved):

mvidner:
In  IBS submit [request #384155](https://build.suse.de/request/show/384155),  @darix says:

> `force_encoding` is not actually changing the encoding of the string but only update the encoding attribute. that might break text. you want `.encode()`  `.encode!()` or so

I know that the string **is** in UTF-8 but misdeclared as binary, so `force_encoding` is exactly what I want. 

I **could** avoid `force_encoding` in favor of `encode` but IMHO that would be pointless:
```console
$ irb
irb(main):001:0> s = "business’ needs"
=> "business’ needs"
irb(main):002:0> s.encoding
=> #<Encoding:UTF-8>
irb(main):003:0> s.force_encoding(Encoding::BINARY)
=> "business\xE2\x80\x99 needs"
irb(main):004:0> s.encoding
=> #<Encoding:ASCII-8BIT>
irb(main):005:0> s.encode(Encoding::UTF_8)
        3: from /usr/bin/irb:11:in `<main>'
        2: from (irb):5
        1: from (irb):5:in `encode'
Encoding::UndefinedConversionError ("\xE2" from ASCII-8BIT to UTF-8)
irb(main):006:0> s.encode(Encoding::UTF_8, Encoding::ASCII_8BIT)
        3: from /usr/bin/irb:11:in `<main>'
        2: from (irb):6
        1: from (irb):6:in `encode'
Encoding::UndefinedConversionError ("\xE2" from ASCII-8BIT to UTF-8)
irb(main):007:0> s.encode(Encoding::UTF_8, Encoding::UTF_8)
=> "business’ needs"
irb(main):008:0> s.encode(Encoding::UTF_8, Encoding::UTF_8).encoding
=> #<Encoding:UTF-8>
```

mvidner:
@darix that the string is in fact UTF-8? Yes:

1. either it comes from SUSE produced metadata
2. or a crazy partner decides that they want to produce an add-on which has crazy-release.rpm encoded in EBCDIC-VOGON. YaST has no way of knowing, so it will interpret it as UTF-8. Fine with me.

PR #661, opened 2025-07-16
Mark strings from libsolv as UTF-8 (bsc#1245555) -> SP7

mvidner:
Merge #660 to the SP7 branch

The code is the same, but the version is increased to be consistent with other SP7 packages


shundhammer (approved):
LGTM

